PR #17, opened 2026-01-14
add project view feature for milestone tracking

chriscanal:
Adds a new "Projects" view that displays GitHub milestones across tracked repositories with progress tracking:

- Add open_issues/closed_issues columns to github_milestones table
- Update milestone sync to fetch and store issue counts from GitHub
- Add /projects/milestones API endpoint for querying milestones
- Create Projects React component with milestone cards grouped by repo
- Show progress bars, due dates, and overdue indicators for milestones
- Add Projects link to dashboard

This enables visibility into project progress across all EquiStamp organization repositories.

Generated with [Claude Code](https://claude.ai/code) via [Happy](https://happy.engineering)


src/memory/api/projects.py
```diff
@@ -0,0 +1,248 @@
...
+# --- Endpoints ---
+
+
+@router.get("/milestones")
+def list_milestones(
+    state: str | None = Query(None, description="Filter by state: 'open', 'closed', or None for all"),
+    repo_filter: list[str] | None = Query(None, description="Filter by repo paths (owner/name)"),
+    include_closed: bool = Query(False, description="Include closed milestones"),
+    user: User = Depends(get_current_user),
+    db: Session = Depends(get_session),
```
mruwnik, line 108: **Quick Ben**: [BLOCKING] **No access control on milestones endpoint.** This endpoint returns all milestones from all active repos to any authenticated user. The rest of the codebase enforces access through User → Person → Team → Project hierarchy (see `access_control.py`), but this endpoint bypasses all of it. Any logged-in user can see every GitHub milestone across all repos.

At minimum, filter by repos/projects the user has access to. If this is intentionally org-wide visibility, it should at least require a specific scope (e.g., `projects` or `admin`).

src/memory/api/projects.py
```diff
@@ -0,0 +1,248 @@
...
+            repos_dict[repo_path].total_open_milestones += 1
+        else:
+            repos_dict[repo_path].total_closed_milestones += 1
+
+    # Build response
+    repos_list = list(repos_dict.values())
+    total_open = sum(r.total_open_milestones for r in repos_list)
+    total_closed = sum(r.total_closed_milestones for r in repos_list)
+
+    return ProjectsOverviewResponse(
```
mruwnik, line 190: **Quick Ben**: [BLOCKING] **Same access control gap on single-milestone endpoint.** `get_milestone` returns any milestone by ID without checking whether the user has access to the repo/project it belongs to. An authenticated user can enumerate milestone IDs to discover milestones in repos they shouldn't see.

src/memory/api/projects.py
```diff
@@ -0,0 +1,248 @@
...
+
+@router.get("/milestones/{milestone_id}")
+def get_milestone(
+    milestone_id: int,
+    user: User = Depends(get_current_user),
+    db: Session = Depends(get_session),
+) -> MilestoneResponse:
+    """Get a single milestone by ID."""
+    result = (
+        db.query(GithubMilestone, GithubRepo)
```
mruwnik, line 207: **Quick Ben**: [BLOCKING] **`list_tracked_repos` also lacks access control.** Returns all active repos to any authenticated user. Same pattern as the milestones endpoints — needs to be scoped to repos the user has access to.

src/memory/api/projects.py
```diff
@@ -0,0 +1,248 @@
...
+
+# --- Helper Functions ---
+
+
+def milestone_to_response(milestone: GithubMilestone, repo: GithubRepo) -> MilestoneResponse:
+    """Convert a GithubMilestone to response model."""
+    open_issues = cast(int, milestone.open_issues) or 0
+    closed_issues = cast(int, milestone.closed_issues) or 0
+    total = open_issues + closed_issues
+    progress = round(100 * closed_issues / max(1, total))
```
mruwnik, line 74: **Quick Ben**: [SUGGESTION] **`cast()` masking potential None values.** `cast(int, milestone.open_issues) or 0` — the `cast()` here is a type-level assertion that doesn't actually convert anything at runtime. If `milestone.open_issues` is `None` (e.g., old rows before the migration backfilled), `cast(int, None)` returns `None` and then `or 0` catches it, so it works by accident. But it's misleading — better to just do `milestone.open_issues or 0` (or `int(milestone.open_issues or 0)` if you want to be explicit).

src/memory/api/projects.py
```diff
@@ -0,0 +1,248 @@
...
+    # Filter by repo paths
+    if repo_filter:
+        # Build OR conditions for each repo path
+        repo_conditions = []
+        for repo_path in repo_filter:
+            if "/" in repo_path:
+                owner, name = repo_path.split("/", 1)
+                repo_conditions.append(
+                    and_(GithubRepo.owner == owner, GithubRepo.name == name)
+                )
```
mruwnik, line 140: **Quick Ben**: [SUGGESTION] **Inline import of `or_` inside the function body.** The `from sqlalchemy import or_` is imported inside the conditional block at line 140. The convention in this codebase (per CLAUDE.md) is to keep imports at the top of the file. `and_` is already imported at the top — `or_` should be alongside it.

src/memory/api/projects.py
```diff
@@ -0,0 +1,248 @@
...
+    return ProjectsOverviewResponse(
+        repos=repos_list,
+        total_repos=len(repos_list),
+        total_open_milestones=total_open,
+        total_closed_milestones=total_closed,
+        last_updated=last_updated.isoformat() if last_updated else None,
+    )
+
+
+@router.get("/milestones/{milestone_id}")
```
mruwnik, line 199: **Quick Ben**: [SUGGESTION] **Inline import of `HTTPException`.** Same issue — `from fastapi import HTTPException` is imported inside the function body at line 199. Should be at the top with the other fastapi imports.

src/memory/common/db/models/sources.py
```diff
@@ -320,6 +324,34 @@ class GithubMilestone(Base):
         Index("github_milestones_due_idx", "due_on"),
     )
```
mruwnik, line 325: **Quick Ben**: [SUGGESTION] **`as_payload()` duplicates the logic in `milestone_to_response()`.** The model now has `as_payload()` which computes `total_issues` and `progress_percent`, but `src/memory/api/projects.py:milestone_to_response()` does the same computation independently. The API endpoint doesn't even use `as_payload()`. Either use `as_payload()` in the endpoint (and add `repo_path`/`repo_name`/`url` to it), or remove `as_payload()` if nothing else calls it. Having two parallel serialization paths for the same model will drift.

src/memory/workers/tasks/github.py
```diff
@@ -55,6 +55,8 @@ def _sync_milestone(
         existing.description = milestone_data["description"]
         existing.state = milestone_data["state"]
         existing.due_on = milestone_data["due_on"]
```
mruwnik, line 57: **Quick Ben**: [IMPORTANT] **Using `.get()` on a TypedDict.** `milestone_data.get("open_issues", 0)` — `GithubMilestoneData` is a `TypedDict` where `open_issues` is a required key (not `NotRequired`). Using `.get()` with a default suggests the field might be absent, but the type system says it's always present. This is either: (a) defensive against older data that predates the field — in which case the TypedDict definition is wrong, or (b) unnecessary — in which case just use `milestone_data["open_issues"]`.

Since the GraphQL parser always populates these fields (with `default=0` in `_extract_nested`), option (b) seems right: use direct access `milestone_data["open_issues"]` for consistency with all the other fields accessed via `[]`.

frontend/src/components/projects/Projects.tsx
```diff
@@ -0,0 +1,361 @@
...
+
+interface RepoSectionProps {
+  repo: RepoMilestones
+  onSelectMilestone: (milestone: Milestone) => void
+  defaultExpanded?: boolean
+}
+
+const RepoSection = ({ repo, onSelectMilestone, defaultExpanded = true }: RepoSectionProps) => {
+  const [expanded, setExpanded] = useState(defaultExpanded)
+
```
mruwnik, line 83: **Quick Ben**: [SUGGESTION] **Hardcoded repo-name-to-client-name heuristic.** `repo.repo_name.replace(/_Issues$/, '').replace(/_/g, ' ')` assumes a naming convention (`METR_Issues` → `METR`) that is specific to one org's conventions. This will show ugly or confusing names for repos that don't follow this pattern (e.g., `memory` → `memory`, `some-repo-name` → `some-repo-name` with dashes intact). Consider just using the repo name directly, or making the display name configurable.

frontend/src/components/projects/Projects.tsx
```diff
@@ -0,0 +1,361 @@
...
+          ))}
+        </div>
+      )}
+    </div>
+  )
+}
+
+const Projects = () => {
+  const { getMilestones, loading, error } = useProjects()
+  const [data, setData] = useState<ProjectsOverview | null>(null)
```
mruwnik, line 119: **Quick Ben**: [IMPORTANT] **Infinite re-render risk from `useCallback` + `useEffect` dependency chain.** `loadData` depends on `getMilestones` and `showClosed`. `getMilestones` is a `useCallback` from `useProjects()` that depends on `apiCall`. If `apiCall` isn't stable (new reference each render), then `getMilestones` gets a new identity each render, which triggers `loadData` recreation, which triggers the `useEffect`, causing an infinite fetch loop.

Check whether `apiCall` from `useAuth()` is stable (wrapped in its own `useCallback` with stable deps). If not, this will loop.

frontend/src/components/projects/Projects.tsx
```diff
@@ -0,0 +1,361 @@
...
+  }, [getMilestones, showClosed])
+
+  useEffect(() => {
+    loadData()
+  }, [loadData])
+
+  return (
+    <div className="min-h-screen bg-slate-50 p-6">
+      <header className="flex items-center gap-4 mb-6 pb-4 border-b border-slate-200">
+        <Link
```
mruwnik, line 135: **Quick Ben**: [NITPICK] **Loading overlay blocks the entire viewport.** The `fixed inset-0` loading indicator covers the whole screen with a white overlay (`bg-white/80`). On initial load this is fine, but on a refresh (clicking the Refresh button), the user loses all visual context. Consider a subtler indicator for refreshes — e.g., a spinner in the header or reducing opacity on the content area rather than a full-screen overlay.

frontend/src/hooks/useProjects.ts
```diff
@@ -0,0 +1,148 @@
...
+
+export interface ProjectsFilters {
+  state?: 'open' | 'closed' | null
+  repoFilter?: string[]
+  includeClosed?: boolean
+}
+
+export const useProjects = () => {
+  const { apiCall } = useAuth()
+  const [loading, setLoading] = useState(false)
```
mruwnik, line 55: **Quick Ben**: [IMPORTANT] **Shared `loading`/`error` state across independent operations.** `getMilestones`, `getMilestone`, and `getTrackedRepos` all share the same `loading` and `error` state. If you call `getMilestones` and then `getTrackedRepos` in parallel, they'll clobber each other's loading/error state. Not a problem today since only `getMilestones` is used, but it's a footgun for anyone who adds a second call. Consider either separate state per method or noting this as a known limitation.

db/migrations/versions/20260114_120000_add_milestone_issue_counts.py
```diff
@@ -0,0 +1,48 @@
...
+Create Date: 2026-01-14
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+# revision identifiers, used by Alembic.
+revision = "20260114_120000"
+down_revision = "20260112_120000"
```
mruwnik, line 17: **Quick Ben**: [IMPORTANT] **`down_revision` may not match the actual migration chain.** This migration declares `down_revision = "20260112_120000"`. If master has gained additional migrations since this branch diverged, this will conflict. The PR already shows as `CONFLICTING` on GitHub — this is likely one of the conflicts. The migration will need to be rebased onto the current head of the migration chain.

mruwnik (commented):
**Quick Ben** *(Claude Opus 4.6)*: ## Review Summary — Request Changes

This PR adds a Projects view for tracking GitHub milestones across repos. The feature is well-structured end-to-end (migration → model → GraphQL fetch → API → React UI), and the UI component is clean and functional. However, there are significant access control gaps that need to be addressed before merging.

**The PR also has merge conflicts** (GitHub shows `CONFLICTING` state) — this needs a rebase onto master.

## Blocking Issues (3)

1. **No access control on `/projects/milestones`** — Returns all milestones from all active repos to any authenticated user, bypassing the User → Person → Team → Project access hierarchy used everywhere else in the codebase.
2. **No access control on `/projects/milestones/{id}`** — Any authenticated user can enumerate milestone IDs to discover milestones in repos they shouldn't see.
3. **No access control on `/projects/repos`** — Returns all active repos to any authenticated user.

All three endpoints need to filter results based on the user's team/project memberships, or at minimum require a specific scope.

## Important Issues (4)

4. **Migration `down_revision` likely conflicts** with current master migration chain — needs rebase.
5. **`.get()` on TypedDict fields** in `_sync_milestone` — inconsistent with the rest of the function which uses `[]` access. The TypedDict says these fields are always present.
6. **Potential infinite re-render loop** in `Projects.tsx` — the `useCallback`/`useEffect` dependency chain may loop if `apiCall` from `useAuth()` isn't referentially stable.
7. **Shared loading/error state** in `useProjects` hook across independent operations — footgun for future callers.

## Suggestions (4)

8. `cast()` usage in `milestone_to_response` is misleading — it doesn't convert at runtime.
9. Inline imports of `or_` and `HTTPException` should be at file top per project conventions.
10. `as_payload()` on the model duplicates logic in `milestone_to_response()` — pick one serialization path.
11. Hardcoded `_Issues` → client name heuristic in the React component is fragile.

## Nitpicks (1)

12. Full-screen loading overlay on refresh is heavy-handed — consider a subtler indicator.

## What Works Well

- Clean separation: GraphQL changes → TypedDict → worker sync → API → React hook → component
- Progress bars, overdue indicators, and the milestone detail modal are well thought out
- Pydantic response models for the API are a good pattern
- The `useProjects` hook interface is clean and idiomatic React

src/memory/api/projects.py
```diff
@@ -0,0 +1,248 @@
...
+    # Build query for all milestones from active repos
+    query = (
+        db.query(GithubMilestone, GithubRepo)
+        .join(GithubRepo, GithubMilestone.repo_id == GithubRepo.id)
+        .join(GithubAccount, GithubRepo.account_id == GithubAccount.id)
+        .filter(
+            GithubRepo.active == True,  # noqa: E712
+            GithubAccount.active == True,  # noqa: E712
+        )
+    )
```
mruwnik, line 123: **Renarin** *(Claude Opus 4.6)*: [BLOCKING] **Missing access control — any authenticated user can read every other user's milestones.**

This query has no `user_id` filter at all. It returns every milestone from every active `GithubAccount` in the entire database, regardless of which user owns the account. Compare to `src/memory/api/github_sources.py:163` for the established pattern:

```python
accounts = db.query(GithubAccount).filter(GithubAccount.user_id == user.id).all()
```

`GithubAccount` already has a `user_id` foreign key, and the join chain `GithubMilestone -> GithubRepo -> GithubAccount` makes filtering easy:

```python
.filter(
    GithubAccount.user_id == user.id,
    GithubRepo.active == True,  # noqa: E712
    GithubAccount.active == True,  # noqa: E712
)
```

This is a real privacy bug — in a multi-tenant setup it leaks milestone titles, descriptions, due dates, and progress numbers from one user's tracked repos to another. Per the project's access-control invariants in `CLAUDE.md`, content access should always be scoped to the requesting user (and admin-scoped users can be opted in explicitly if that's wanted).

Same issue applies to `get_milestone` (line ~204) and `list_tracked_repos` (line ~227) below — neither filters by `user_id` either.

src/memory/api/projects.py
```diff
@@ -0,0 +1,248 @@
...
+    """Get a single milestone by ID."""
+    result = (
+        db.query(GithubMilestone, GithubRepo)
+        .join(GithubRepo, GithubMilestone.repo_id == GithubRepo.id)
+        .filter(GithubMilestone.id == milestone_id)
+        .first()
+    )
+
+    if not result:
+        from fastapi import HTTPException
```
mruwnik, line 214: **Renarin** *(Claude Opus 4.6)*: [BLOCKING] No access control here either — any authenticated user can fetch any milestone by ID just by guessing/iterating IDs. Add the same `GithubAccount.user_id == user.id` filter (joining through `GithubRepo`) that you'd apply in `list_milestones`. As written, this also won't 404 properly when the milestone exists but belongs to another user — it will leak the data.

src/memory/api/projects.py
```diff
@@ -0,0 +1,248 @@
...
+) -> list[dict]:
+    """List all tracked repos that can have milestones."""
+    repos = (
+        db.query(GithubRepo)
+        .join(GithubAccount)
+        .filter(
+            GithubRepo.active == True,  # noqa: E712
+            GithubAccount.active == True,  # noqa: E712
+        )
+        .order_by(GithubRepo.owner, GithubRepo.name)
```
mruwnik, line 235: **Renarin** *(Claude Opus 4.6)*: [BLOCKING] Same access-control gap — needs `GithubAccount.user_id == user.id` filter so users only see repos belonging to their own account(s).

src/memory/api/projects.py
```diff
@@ -0,0 +1,248 @@
...
+        # Build OR conditions for each repo path
+        repo_conditions = []
+        for repo_path in repo_filter:
+            if "/" in repo_path:
+                owner, name = repo_path.split("/", 1)
+                repo_conditions.append(
+                    and_(GithubRepo.owner == owner, GithubRepo.name == name)
+                )
+        if repo_conditions:
+            from sqlalchemy import or_
```
mruwnik, line 142: **Renarin** *(Claude Opus 4.6)*: [IMPORTANT] Imports inside the function — `from sqlalchemy import or_` here and `from fastapi import HTTPException` at line ~210. Per the project's style guide (see `/workspace/memory/.claude/CLAUDE.md` and the global guide), "Imports should always be at top of the file, unless you have a really good reason to do otherwise." There's no reason here; both should be at the top alongside the other `sqlalchemy` and `fastapi` imports. Bonus: `HTTPException` raised mid-function would shadow nicely if you ever introduced linting around top-level imports.

src/memory/api/projects.py
```diff
@@ -0,0 +1,248 @@
...
+    total = open_issues + closed_issues
+    progress = round(100 * closed_issues / max(1, total))
+
+    return MilestoneResponse(
+        id=cast(int, milestone.id),
+        repo_path=repo.repo_path,
+        repo_name=cast(str, repo.name),
+        number=cast(int, milestone.number),
+        title=cast(str, milestone.title),
+        description=cast(str | None, milestone.description),
```
mruwnik, line 82: **Renarin** *(Claude Opus 4.6)*: [IMPORTANT] The `cast(int, ...) or 0` pattern is doubly problematic:

1. `typing.cast` is purely a static-typing hint and does **nothing** at runtime — `cast(int, milestone.open_issues)` returns whatever `milestone.open_issues` is, including `None` if it were ever `None`.
2. The columns `open_issues` and `closed_issues` are declared `nullable=False, server_default="0"` (in `sources.py`), so they cannot be `None` for a persisted row. The `or 0` is dead defense.

If the type checker is complaining about column attribute types, that's a SQLAlchemy stub problem and the fix is to type the columns properly via `Mapped[int]` (or `mapped_column`) rather than papering over it with `cast` everywhere. Right now this file uses `cast` 9 times — that's a code smell. The same pattern is repeated in `github_sources.py`, but that doesn't make it good; this PR is a chance to do it right (or at least not multiply the problem).

Minimal fix here: drop the `cast(int, ...)` calls and the `or 0` — trust the schema.

src/memory/api/projects.py
```diff
@@ -0,0 +1,248 @@
...
+    return [
+        {
+            "id": repo.id,
+            "repo_path": repo.repo_path,
+            "owner": repo.owner,
+            "name": repo.name,
+            "last_sync_at": repo.last_sync_at.isoformat() if repo.last_sync_at else None,
+        }
+        for repo in repos
+    ]
```
mruwnik, line 248: **Renarin** *(Claude Opus 4.6)*: [BLOCKING] **No tests included for any of the new endpoints.**

`tests/memory/api/` has tests for every other API module: `test_github_sources.py`, `test_polls.py`, `test_email_accounts.py`, etc. This PR adds three endpoints (`/projects/milestones`, `/projects/milestones/{id}`, `/projects/repos`) and zero tests.

A `test_projects.py` here would have caught:
- The access-control gaps flagged above (a test with two users + one repo each is the standard pattern).
- The progress-percent edge case at `total == 0` (silent reliance on `max(1, total)`).
- The `repo_filter` parsing of paths without a `/`.

The existing fixtures in `tests/conftest.py` (`db_session`, `admin_user`, `regular_user`, `admin_session`, `user_session`) make this cheap to write. Please add a test file before merging — the access control alone justifies it.

src/memory/api/projects.py
```diff
@@ -0,0 +1,248 @@
...
+        GithubRepo.owner,
+        GithubRepo.name,
+        GithubMilestone.title,
+    )
+
+    results = query.all()
+
+    # Group by repo
+    repos_dict: dict[str, RepoMilestonesResponse] = {}
+    last_updated = None
```
mruwnik, line 158: **Renarin** *(Claude Opus 4.6)*: [SUGGESTION] If `repo_filter` is provided but every entry is malformed (no `/`), `repo_conditions` is empty and the `or_(*[])` block is silently skipped — the request acts as if no filter were given. That probably surprises clients that pass a typo'd filter and get the full list back. Either:
- 400 if `repo_filter` is non-empty but yields zero valid conditions, or
- match nothing (`query = query.filter(False)`).

Silent fall-through to "all results" is the worst option.

src/memory/common/db/models/sources.py
```diff
@@ -320,6 +324,34 @@ class GithubMilestone(Base):
...
+            "repo_id": self.repo_id,
+            "github_id": self.github_id,
+            "number": self.number,
+            "title": self.title,
+            "description": self.description,
+            "state": self.state,
+            "due_on": self.due_on.isoformat() if self.due_on else None,
+            "open_issues": self.open_issues,
+            "closed_issues": self.closed_issues,
+            "total_issues": (self.open_issues or 0) + (self.closed_issues or 0),
```
mruwnik, line 340: **Renarin** *(Claude Opus 4.6)*: [NITPICK] `as_payload` appears to be **dead code** — the new API in `src/memory/api/projects.py` builds responses via `milestone_to_response` instead, and grep wouldn't show any other caller. It also duplicates the `progress_percent` calculation (`round(100 * closed / max(1, open + closed))`) inline.

Either:
- Drop `as_payload` entirely and let `milestone_to_response` be the single source of truth, or
- Use `as_payload` as a `model.dict()`-like helper and have `milestone_to_response` just pass through.

Right now you have two places to update if the schema changes, with a non-zero chance they'll drift.

src/memory/common/github/issues.py
```diff
@@ -524,6 +524,8 @@ def fetch_pr_diff(
         createdAt
         updatedAt
         closedAt
+        issues(first: 0, states: [OPEN]) { totalCount }
```
mruwnik, line 527: **Renarin** *(Claude Opus 4.6)*: [IMPORTANT] `first: 0` works on GitHub's GraphQL API today (it returns the connection without nodes), but it's surprising and arguably abusing the schema — the `first` argument is documented as "Returns the first n elements from the list" with the implicit expectation `n >= 1`. Some GraphQL servers reject `first: 0` outright and the GitHub schema has been progressively tightened over the years.

Safer pattern: keep the connection field but request only `totalCount`:

```graphql
issues(states: [OPEN]) { totalCount }
closedIssues: issues(states: [CLOSED]) { totalCount }
```

If GitHub requires `first` here for paginated connections, use `first: 1` and just don't read `nodes`. The cost is one extra `id` round-trip; the benefit is not depending on the server tolerating `first: 0`.

src/memory/workers/tasks/github.py
```diff
@@ -55,6 +55,8 @@ def _sync_milestone(
...
+        existing.open_issues = milestone_data.get("open_issues", 0)
+        existing.closed_issues = milestone_data.get("closed_issues", 0)
```
mruwnik, line 59: **Renarin** *(Claude Opus 4.6)*: [NITPICK] `milestone_data.get("open_issues", 0)` is defensive against a key that the `GithubMilestoneData` TypedDict (after this PR) declares as required (`open_issues: int`). Either the TypedDict is wrong (and these should be `NotRequired[int]`), or the `.get(..., 0)` fallback is dead. Pick one.

If the concern is "older code paths might construct `GithubMilestoneData` without these fields", that's exactly the kind of thing the type system should catch — and `mypy` will flag missing required keys at the construction site, not here.

frontend/src/hooks/useProjects.ts
```diff
@@ -0,0 +1,148 @@
...
+  last_sync_at: string | null
+}
+
+export interface ProjectsFilters {
+  state?: 'open' | 'closed' | null
+  repoFilter?: string[]
+  includeClosed?: boolean
+}
+
+export const useProjects = () => {
```
mruwnik, line 53: **Renarin** *(Claude Opus 4.6)*: [IMPORTANT] Three independent async operations (`getMilestones`, `getMilestone`, `getTrackedRepos`) share a single `loading` boolean and a single `error` string. If a consumer triggers two of them concurrently (or in quick succession), the latter overwrites the former's state and the UI becomes inconsistent — e.g. one finishes, sets `loading=false`, while the other is still in-flight, so the spinner disappears prematurely.

Standard fix: separate state per call, or return state from each call rather than tracking it on the hook. Pattern that works here:

```ts
const useProjects = () => {
  const { apiCall } = useAuth();
  // call-specific state, or just have callers manage their own
  ...
};
```

Right now `Projects.tsx` only uses `getMilestones`, so this isn't observable in this PR — but the hook is exported as a general-purpose API and the next caller will hit the bug.

frontend/src/components/projects/Projects.tsx
mruwnik, line 119: **Renarin** *(Claude Opus 4.6)*: [IMPORTANT] Race condition in the load effect. If the user toggles "Show closed" rapidly, two `loadData()` calls are in flight; the slower one resolves last and clobbers `data` even if it represents the *older* `showClosed` state.

Standard fix using a cancellation flag:

```tsx
useEffect(() => {
  let cancelled = false;
  (async () => {
    const result = await getMilestones({ includeClosed: showClosed });
    if (!cancelled) setData(result);
  })();
  return () => { cancelled = true; };
}, [getMilestones, showClosed]);
```

Or use an `AbortController` if `apiCall` supports `signal`. Right now the only thing saving you is that toggling a checkbox is unlikely to land mid-fetch — but the same effect runs on every reference change.

frontend/src/components/projects/Projects.tsx
```diff
@@ -0,0 +1,361 @@
...
+  defaultExpanded?: boolean
+}
+
+const RepoSection = ({ repo, onSelectMilestone, defaultExpanded = true }: RepoSectionProps) => {
+  const [expanded, setExpanded] = useState(defaultExpanded)
+
+  // Extract client name from repo name (e.g., "METR_Issues" -> "METR")
+  const clientName = repo.repo_name.replace(/_Issues$/, '').replace(/_/g, ' ')
+
+  return (
```
mruwnik, line 87: **Renarin** *(Claude Opus 4.6)*: [IMPORTANT] Hard-coding the `_Issues` suffix and underscore-to-space conversion couples this UI to one specific naming convention — apparently EquiStamp's `<CLIENT>_Issues` pattern. For repos like `memory`, `cairn`, or anything not following this pattern, the user sees the bare repo name (which is fine) but for `Foo_Bar` it's silently rendered as `Foo Bar` — which may or may not be a real client name.

If the goal is "show a friendly client/project name", model that explicitly: add a `display_name` column to `GithubRepo` (or a tag/alias) and let users configure it on the Sources page. As-is, this is a hack that works in one org and silently does the wrong thing elsewhere.

At minimum, add a comment explaining the assumption — and consider making this regex configurable.

frontend/src/components/projects/Projects.tsx
```diff
@@ -0,0 +1,361 @@
...
+        <div className="text-center py-12">
+          <div className="text-slate-400 text-lg mb-2">No milestones found</div>
+          <p className="text-slate-500 text-sm">
+            Configure GitHub repos to track in the Sources page
+          </p>
+          <Link
+            to="/ui/sources"
+            className="inline-block mt-4 text-primary hover:underline"
+          >
+            Go to Sources
```
mruwnik, line 222: **Renarin** *(Claude Opus 4.6)*: [SUGGESTION] Modal accessibility gaps:
- No `role="dialog"` / `aria-modal="true"` on the dialog element.
- No focus management — focus stays wherever it was (typically the milestone card), and tab order leaks out of the modal.
- No `Escape` key handler — only the backdrop click and the × button can close it.
- The close button has no `aria-label="Close"` — screenreaders just hear "×".

These aren't strictly blocking, but they're cheap to add and the rest of the codebase appears to use plain Tailwind divs for modals so this is the moment to set a better baseline. Bonus: a tiny `useEffect` listening for `keydown` / `Escape` is ~5 lines.

frontend/src/components/projects/Projects.tsx
```diff
@@ -0,0 +1,361 @@
...
+        </div>
+        <div className="h-2 bg-slate-100 rounded-full overflow-hidden">
+          <div
+            className={`h-full transition-all ${
+              milestone.progress_percent === 100
+                ? 'bg-green-500'
+                : milestone.progress_percent > 50
+                  ? 'bg-primary'
+                  : 'bg-yellow-500'
+            }`}
```
mruwnik, line 52: **Renarin** *(Claude Opus 4.6)*: [NITPICK] A milestone with `total_issues == 0` will display:
- Progress bar: yellow (because `progress_percent === 0`, falls into the `<= 50%` branch)
- "0% complete" with "0/0 issues"

That's misleading — there's nothing to be 0%-complete on. Two cleaner options:
- Render the progress bar as gray/empty when `total_issues === 0` and show "No issues" instead of "0% complete".
- Hide the progress bar entirely for empty milestones.

Same applies to the modal at line ~239.

frontend/src/components/projects/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default as Projects } from './Projects'
+export { default } from './Projects'
```
mruwnik, line 2: **Renarin** *(Claude Opus 4.6)*: [NITPICK] Both lines re-export the same default. The `export { default as Projects } from './Projects'` is what `frontend/src/components/index.js` consumes; the second line `export { default } from './Projects'` is unused. Drop it for clarity, or keep just one — having both invites confusion.

db/migrations/versions/20260114_120000_add_milestone_issue_counts.py
mruwnik, line 17: **Renarin** *(Claude Opus 4.6)*: [SUGGESTION] `down_revision = "20260112_120000"` — the PR has been open since January 2026 and is currently `CONFLICTING` with master. Verify on rebase that the current head of the migration chain hasn't moved past `20260112_120000`. If so, this `down_revision` will need updating; otherwise alembic will refuse to run with multiple heads. (Today the head is indeed `20260112_120000`, so it's fine for now — just flag for the rebase.)

frontend/src/hooks/useProjects.ts
```diff
@@ -0,0 +1,148 @@
...
+
+export interface ProjectsOverview {
+  repos: RepoMilestones[]
+  total_repos: number
+  total_open_milestones: number
+  total_closed_milestones: number
+  last_updated: string | null
+}
+
+export interface TrackedRepo {
```
mruwnik, line 39: **Renarin** *(Claude Opus 4.6)*: [NITPICK] The `state` field on `ProjectsFilters` is sent as `params.append('state', filters.state)` (line ~64), but the backend currently treats `state` and `include_closed` separately. If you set both — e.g. `state='closed'` with `includeClosed: false` — the server uses `state` and ignores `include_closed`, which is a fine resolution but worth documenting. Consider collapsing the API to a single `state: 'open' | 'closed' | 'all'` parameter; the current shape is two ways to express the same intent and the dual semantics are easy to get wrong.

src/memory/api/projects.py
```diff
@@ -0,0 +1,248 @@
...
+    milestone, repo = result
+    return milestone_to_response(milestone, repo)
+
+
+@router.get("/repos")
+def list_tracked_repos(
+    user: User = Depends(get_current_user),
+    db: Session = Depends(get_session),
+) -> list[dict]:
+    """List all tracked repos that can have milestones."""
```
mruwnik, line 227: **Renarin** *(Claude Opus 4.6)*: [NITPICK] Return type is `list[dict]`, but every other endpoint in this module uses a typed `BaseModel`. Define a `TrackedRepoResponse` (you already have an interface for it on the frontend at `useProjects.ts`'s `TrackedRepo`) and use it as the return type. Benefits: OpenAPI schema is correct, response validation happens, and the response shape is documented in code rather than implicit.

mruwnik (commented):
**Renarin** *(Claude Opus 4.6)*: ## Review Summary

**Decision: Request changes.** The feature itself is reasonable and the UI is well-styled, but there's a multi-tenant access-control gap on every new endpoint that has to be fixed before merging. There are also no tests for the new API, which would have caught it.

## Blocking Issues

1. **`/projects/milestones`, `/projects/milestones/{id}`, `/projects/repos` have no `user_id` filter.** Any authenticated user sees every other user's tracked repos and milestone titles/descriptions/due dates/progress. The fix is straightforward — add `GithubAccount.user_id == user.id` to each query, mirroring the established pattern in `github_sources.py`.
2. **No tests added.** Every other API in `src/memory/api/` has a corresponding `tests/memory/api/test_*.py`. Add a `test_projects.py` covering: access control (two users + cross-account read attempts), state filtering, `repo_filter` parsing, progress-percent edge cases (0/0, 0/n, n/0).

## Important Issues

- `cast(int, milestone.X) or 0` pattern in `milestone_to_response` — `cast` is a no-op at runtime; the columns are non-nullable; both layers are unnecessary. Same `cast` overuse in `as_payload`.
- Imports inside functions (`or_`, `HTTPException`) — violates project style guide. Move to module top.
- `useProjects` hook shares one `loading`/`error` between three async ops; concurrent calls clobber each other's state.
- `Projects.tsx` `loadData` effect has no race-condition guard — rapid `showClosed` toggles can land out of order and persist stale data.
- Hard-coded `_Issues` suffix stripping for "client name" — couples this UI to one org's repo-naming convention; for any other repo it silently does the wrong/right thing depending on the name.
- GraphQL `first: 0` in `issues.py` works today but is fragile — prefer `first: 1` or rely on `totalCount` without `first`.

## Suggestions

- `repo_filter` with all-malformed entries silently returns the full list (worst-of-both-worlds). 400 or filter-to-empty.
- Modal has zero accessibility attributes (no `role="dialog"`, no Escape handler, no focus trap, no `aria-label` on close button).
- "0/0 issues" milestones render as 0% with a yellow bar — UX nit, but worth handling.
- `list_tracked_repos` returns `list[dict]`; should use a typed Pydantic model for consistency with the rest of the file.
- `as_payload` on `GithubMilestone` appears to be dead code that duplicates `milestone_to_response` logic — remove or unify.
- `frontend/src/components/projects/index.ts` exports `default` and `Projects` redundantly.
- `state` query param + `include_closed` query param overlap; consider a single `state: open | closed | all`.

## Nitpicks

- Migration `down_revision = "20260112_120000"` is correct today but verify on rebase (PR is `CONFLICTING`).
- `milestone_data.get("open_issues", 0)` is defensive against a TypedDict-required field — pick one or the other.

## What Works Well

- Clean separation between API layer, hook layer, and component layer.
- Pydantic response models for the milestone endpoints (just extend the same pattern to `/repos`).
- Good use of `useCallback` for `loadData` and stable references in the hook.
- Migration is symmetric (reversible) and uses sensible server defaults.
- The progress-bar / due-date color coding is a thoughtful UX touch — overdue red, due-soon yellow, on-track default.
- The `_parse_milestone_graphql` change correctly threads the new fields through `_extract_nested` with sensible defaults.

mruwnik:
done differently now

Issue #26, opened 2026-01-19
Multiple API keys


mruwnik:
Currently users have various API keys, but these tend to be in the user table or spread around other places. There should be a new table for API keys with at least an expiry date and a key type (e.g. discord, google, internal server). All api keys in the user table should be migrated to this new table, and the user model should have a list of api keys (a relationship). Places that check api keys should properly handle a given user having multiple keys.
Potentially the OAuth tables can be chanegd to use this, but only do so it it will make things clearer.
There should also be one time API keys. These are key which as soon as they are used, they get deleted (or marked as inactive or whatever). The `get_user` MCP tool should accept an optional argument to generate one of these, which will then be returned. This is so MCP clients can have a way of running one off calls against the api. API endpoints should be able to specify what kinds of tokens they accept (the default being all).

PR #42, opened 2026-01-22
Add multiple API keys support with expiration and one-time keys (#26)

mruwnik:
## Summary

Implements centralized API key management for issue #26:

- **New `APIKey` model** with key types (discord, google, internal, github, mcp, one_time), expiration dates, one-time use keys, and custom scopes
- **Database migration** to create `api_keys` table and migrate existing keys
- **Updated authentication** to check new table first with proper fallback logic (blocking revoked key bypass)
- **API endpoints** for key management (`/users/me/api-keys`, `/{user_id}/api-keys`)
- **MCP tool update** - `get_user()` can generate one-time keys for client operations
- **`require_key_types()`** dependency to restrict endpoints to specific key types

### Key Features

1. Multiple keys per user with different types and expiration
2. One-time keys deleted after first use (for secure client operations)
3. Per-key scope overrides
4. Backwards-compatible with legacy `User.api_key` field
5. Proper timezone handling for expiration checks
6. Key type validation on API inputs

## Test plan

- [x] All existing auth tests pass (49 passed)
- [x] New tests for APIKey model methods
- [x] New tests for authentication with key type restrictions
- [x] Tests for one-time key deletion
- [x] Tests for revoked key rejection
- [ ] Integration test with database (requires PostgreSQL)

### DIFF
--- a/src/memory/common/db/models/users.py
+++ b/src/memory/common/db/models/users.py
@@ -241,10 +241,8 @@ class APIKey(Base):
     key_type: Mapped[str] = mapped_column(
         String, nullable=False, default=APIKeyType.INTERNAL
     )
-    # Scopes override - RESERVED FOR FUTURE USE
-    # Currently, authentication uses the user's scopes directly.
-    # This field is intended for future per-key scope restrictions but is not
-    # yet enforced. If None, uses user's default scopes (current behavior).
+    # Scopes override - if set, these scopes are used instead of the user's scopes.
+    # If None, authentication falls back to the user's default scopes.
     scopes: Mapped[list[str] | None] = mapped_column(ARRAY(String), nullable=True)
     created_at: Mapped[datetime | None] = mapped_column(
         DateTime, server_default=func.now()


src/memory/api/auth.py
```diff
@@ -98,44 +99,104 @@ def authenticate_bot(api_key: str, db: DBSession) -> BotUser | None:
     """Authenticate a bot by API key.
 
     Uses constant-time comparison to prevent timing attacks.
+    First checks the new api_keys table, then falls back to legacy User.api_key.
```
mruwnik: **Claude Opus 4.5**: **Performance concern**: `db.query(APIKey).all()` loads ALL API keys into memory for every authentication request. This becomes O(n) in the number of total API keys in the system.

Consider using a query with a filter first:
```python
key_record = db.query(APIKey).filter(APIKey.key == api_key).first()
if key_record:
    if not key_record.is_valid():
        return None
    ...
```

While this technically reveals timing information about key existence vs validity, the current approach trades security for a significant performance penalty. The index on `api_keys.key` would make the filtered query O(1) via index lookup instead of O(n).

src/memory/api/auth.py
```diff
@@ -98,44 +99,104 @@ def authenticate_bot(api_key: str, db: DBSession) -> BotUser | None:
...
+    if allowed_key_types and "internal" not in allowed_key_types:
+        return None, None
+
     users = db.query(User).filter(User.api_key.isnot(None)).all()
     for user in users:
         if user.api_key and secrets.compare_digest(user.api_key, api_key):
-            return user
-    return None
+            return user, None
+    return None, None
```
mruwnik: **Claude Opus 4.5**: **Issue**: The `handle_api_key_use` function commits within itself, which could cause issues if called within a larger transaction context.

The caller might want to do additional work before committing, especially if the one-time key deletion should be atomic with other operations. Consider:
1. Removing the `db.commit()` and letting the caller manage the transaction, OR
2. At minimum, documenting that this function commits

src/memory/api/MCP/servers/meta.py
```diff
@@ -141,10 +143,33 @@ async def get_current_time() -> dict:
...
+        result = _get_current_user(session)
+
+        if generate_one_time_key and result.get("authenticated"):
+            access_token = get_access_token()
+            if access_token:
+                user_session = session.get(UserSession, access_token.token)
+                if user_session and user_session.user:
+                    # Create a one-time API key
+                    one_time_key = APIKey.create(
+                        user_id=user_session.user.id,
```
mruwnik: **Claude Opus 4.5**: **Good**: The one-time key generation is nicely integrated into the existing `get_user()` flow with proper guards checking authentication status and user session validity.

src/memory/common/db/models/users.py
```diff
@@ -193,6 +196,109 @@ def create_with_api_key(
...
+        scopes: list[str] | None = None,
+        is_one_time: bool = False,
+        expires_at: datetime | None = None,
+        prefix: str | None = None,
+    ) -> "APIKey":
+        """Create a new API key for a user."""
+        if prefix is None:
+            prefix = "ot" if is_one_time else key_type
+        return cls(
+            user_id=user_id,
```
mruwnik: **Claude Opus 4.5**: **Good defensive coding**: The timezone handling in `is_valid()` correctly handles both tz-aware and tz-naive datetimes, which is important since PostgreSQL can return either depending on column type.

src/memory/api/auth.py
```diff
@@ -98,44 +106,132 @@ def authenticate_bot(api_key: str, db: DBSession) -> BotUser | None:
...
+        # Key found in new table - don't fall back to legacy even if invalid/revoked
+        if not matched_key.is_valid():
+            return None
+        user = matched_key.user
+        if user.user_type == "bot":
+            handle_api_key_use(matched_key, db)
+            return cast(BotUser, user)
+        return None
+
+    # Fall back to legacy api_key on User model for backwards compatibility
```
mruwnik: nope

db/migrations/versions/20260122_150000_add_api_keys_table.py
```diff
@@ -0,0 +1,78 @@
...
+    # Create the api_keys table
+    op.create_table(
+        "api_keys",
+        sa.Column("id", sa.Integer(), autoincrement=True, nullable=False),
+        sa.Column("user_id", sa.Integer(), nullable=False),
+        sa.Column("key", sa.String(), nullable=False),
+        sa.Column("name", sa.String(), nullable=True),
+        sa.Column("key_type", sa.String(), nullable=False, server_default="internal"),
+        sa.Column("scopes", postgresql.ARRAY(sa.String()), nullable=True),
+        sa.Column("is_one_time", sa.Boolean(), nullable=False, server_default="false"),
```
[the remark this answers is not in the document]
mruwnik: **claude** *(Claude Opus 4.5)*: Good point! Removed the `is_one_time` column. Now one-time keys are identified by `key_type="one_time"` and the `is_one_time` property is derived from that.

src/memory/api/users.py
```diff
@@ -239,8 +251,14 @@ def regenerate_api_key(
     user_id: int,
     user: User = Depends(get_current_user),
     db: Session = Depends(get_session),
-) -> ApiKeyResponse:
-    """Regenerate API key for a user. Admins can regenerate any user's key, others only their own."""
+) -> LegacyApiKeyResponse:
```
mruwnik: no legacy, please

db/migrations/versions/20260122_150000_add_api_keys_table.py
```diff
@@ -0,0 +1,78 @@
...
+        sa.Column("revoked", sa.Boolean(), nullable=False, server_default="false"),
+        sa.PrimaryKeyConstraint("id"),
+        sa.ForeignKeyConstraint(["user_id"], ["users.id"], ondelete="CASCADE"),
+        sa.UniqueConstraint("key", name="unique_api_key"),
+    )
+    op.create_index("idx_api_keys_user_id", "api_keys", ["user_id"])
+    op.create_index("idx_api_keys_key", "api_keys", ["key"])
+    op.create_index("idx_api_keys_key_type", "api_keys", ["key_type"])
+
+    # Migrate existing api_key values from users table
```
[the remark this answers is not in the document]
mruwnik: **claude** *(Claude Opus 4.5)*: After investigation, skipping this migration for now due to architectural constraints:

- **DiscordBot.token_encrypted**: Not user-scoped (many-to-many with users via discord_bot_users table). The token belongs to the bot, not a specific user, so it doesn't fit the api_keys model which requires user_id.

- **GithubAccount.access_token**: Has related auth fields (auth_type, app_id, installation_id, private_key for GitHub App auth) that would need to stay together.

These are external service credentials (for calling external APIs) rather than internal authentication tokens (for authenticating to our system). Happy to revisit if you want to design a unified credentials table that handles these different cases.

src/memory/common/db/models/users.py
```diff
@@ -193,6 +196,133 @@ def create_with_api_key(
...
+    )
+    key: Mapped[str] = mapped_column(String, nullable=False, unique=True)
+    name: Mapped[str | None] = mapped_column(String, nullable=True)
+    key_type: Mapped[str] = mapped_column(
+        String, nullable=False, default=APIKeyType.INTERNAL
+    )
+    # Scopes override - RESERVED FOR FUTURE USE
+    # Currently, authentication uses the user's scopes directly.
+    # This field is intended for future per-key scope restrictions but is not
+    # yet enforced. If None, uses user's default scopes (current behavior).
```
mruwnik: can you have the code check this now? Even if it's not set anywhere?

PR #74, opened 2026-05-01
Deep audit: 44 fixes for security, a11y, correctness, and types

mruwnik:
Output of a `/deep-audit` run by the audit-caladan team. 44 commits, ~52 files changed, ~1377 net lines added across the four buckets below.

### Security (the bulk of the PR — ~25 fixes)
**Path traversal:** validated `differ_path` and WebSocket `ws_subpath` in `cloud_claude.py`, validated `filename` in MCP report upsert (matches the existing `notes.py` pattern).
**Cross-user data exposure:** added per-item access control to `/files/{path}` (was open to any authenticated user); per-user isolation for notes (`note_files` filtered by `creator_id`); ownership check in MCP `fetch_file`; ownership check in `_load_attachment` (email exfiltration); applied access filter to `_fetch_chunks_by_title` (closes a prompt-injection exfiltration path); per-user isolation for Discord channel/server endpoints.
**Auth/session integrity:** MCP session tokens now check `expires_at` (mirrors `auth.py`); `fetch_user_by_token` validates API key state and consumes one-time keys; `get_or_create_session` checks ownership before returning; `_CLAUDE_SESSION_PATTERN` whitelist updated to cover snapshot- and env-based sessions (was 401-ing legitimate WebSockets).
**Authorization gaps:** `list_items`/`count_items` now apply the same `creator_id` access condition as `user_can_access` (users can see their own unprojected items); `teams.upsert` and `team_add_member` validate caller's team role (closes privilege escalation); `tidbit_add`/`tidbit_update` validate target project membership; `resolve_bot_id` delegates to ownership-checking `resolve_bot`; GitHub `list_issues`/`list_milestones` and `create_repo_project`/`create_milestone_project` apply per-user access checks; API key creation cannot grant scopes the caller doesn't already hold.
**OAuth hardening:** dynamic client registration restricted to `OAUTH_REDIRECT_URI_ALLOWLIST` (default localhost-only); auth code exchange validates `client_id` binding (RFC 6749 §4.1.3); token expiry uses UTC; bad state on login returns 200 form-with-error instead of 500.
**Header/CSP injection:** sanitize email headers (CRLF injection); validate CSP `allowed_connect_urls` against header-special chars at both write and serve layers.
**Critical: BM25 phantom view** — `bm25.apply_access_filter` referenced a `source_item_access_view` that exists in **no migration**. On any fresh deploy, BM25 access filtering silently failed (degraded, not bypass). Rewritten to use `SourceItem` columns directly, mirroring `core.apply_access_control_to_query`.

### Accessibility (frontend)
- `SelectableTags`: `<span onClick>` → `<button aria-pressed>`
- Fleet rows: `<tr>` made keyboard-focusable with `role="button"`, tabIndex, Enter/Space handler
- `UserMenu`: full ARIA Menu Button pattern (Escape-to-close, arrow nav, `aria-expanded`)
- `PersonFormModal`: `role="dialog"`, `aria-modal`, `aria-labelledby`, focus trap
- `Modal` / `ConfirmDialog` / `StatusBadge` ARIA semantics (sources/shared)
- `PollGrid` availability selector keyboard- and screen-reader accessible

### Backend correctness
- Removed global `task_autoretry_for=(Exception,)` (was retrying programming errors and could spawn 4× Claude containers)
- `validate_slot` normalizes tz-naive client datetimes (was 500-ing on common ISO strings without `Z`)
- Consolidated `has_admin_scope` in `access_control.py`; removed the divergent copy from `auth.py`
- Fixed `get_current_user()` returning `authenticated=True` when the user lookup failed
- Fixed `load_access_token` `AssertionError` when `UserSession` has no `oauth_state_id`

### Types and code health
- `UserLike`/`SourceItemLike` protocols tightened from `Any` to concrete types
- `ImageContent.image` / `ToolUseContent.input` annotations match runtime defaults (Optional)
- Removed dead `generate_negative_project_id`
- Removed stray `if`-conditionals from tests (per CLAUDE.md)
- Removed unused `UserSession` import in MCP base
- Several new tests: `test_polls.py`, `test_users_api.py`, `test_cloud_claude.py`, plus updates in `test_teams.py`

## What this PR does NOT do

- **No internal review pass.** Phase 3 of the audit (a fresh-eyes review of every fix in differ-review) was skipped due to an agent-pane resource constraint. The validator (`Jefri`) also did not get to validate the in-flight `in_review` tasks for the same reason. **Reviewer should treat this as an unreviewed-by-bot batch** and read carefully — particularly the auth/access changes and the BM25 rewrite.
- **MEDIUM and LOW findings deferred.** ~75 backlog items remain on the audit kanban: more datetime tz cleanup, mutable-default cleanup, layering refactors (e.g., `cloud_claude.py` decomposition), additional access-control consistency, performance fixes (`summarize_stale_sessions`, `_fetch_chunks_by_title` ORM load). A follow-up audit iteration is appropriate.
- **No automated full-suite test run.** Workers ran targeted tests; a CI run on this branch is the right place to catch regressions.

## Test plan
- [ ] `pytest` full suite green
- [ ] Spot-check the BM25 access-filter rewrite against a real PostgreSQL session (the filter changed from view-based to column-based; defense-in-depth invariant for non-admin users)
- [ ] Spot-check `/files/{path}` and the new MCP `fetch_file` ownership checks against a multi-user scenario
- [ ] Verify auth middleware whitelist still passes legitimate WebSocket traffic for `u{id}-s{snap}-{hex}` and `u{id}-x-{hex}` session URLs
- [ ] Verify OAuth `client_id` binding doesn't break the existing Cursor/Claude-Code MCP clients
- [ ] A11y smoke test: tab through search filters, fleet table, dialogs, user menu, and poll grid using keyboard only


src/memory/api/discord.py
```diff
@@ -440,18 +440,26 @@ def list_bot_users(
 # --- Server Endpoints ---
 
 
+def user_can_manage_server(user: User, db: Session, server: DiscordServer) -> bool:
+    """Check that the user has a bot linked to this server.
+
+    Returns True when:
+    - The server has no owning bot (legacy rows — allow for backwards compat)
+    - The server's bot_id belongs to a bot the user is authorised for
+    """
```
mruwnik, line 454: **Ellie** *(Claude Opus 4.7)*: **SECURITY (HIGH): The `bot_id IS NULL` "legacy backwards-compat" branch is an open back door.**

```python
def user_can_manage_server(user: User, db: Session, server: DiscordServer) -> bool:
    if server.bot_id is None:
        return True   # ← any authenticated user can manage these
    ...
```

Combined with the new migration adding `bot_id` as nullable with **no backfill**:
```python
op.add_column("discord_servers", sa.Column("bot_id", sa.BigInteger(), ..., nullable=True))
```
…every existing `discord_servers` row immediately after migration has `bot_id = NULL`, meaning *every* `discord`-scoped user can update collection settings on every pre-existing server until the bot reconnects and `ensure_server` back-fills it. For a security-hardening PR this is the opposite of the stated intent (per-user isolation).

Two recommended fixes (not mutually exclusive):
1. Backfill `bot_id` in the migration. If there is exactly one bot per deployment, set every server's `bot_id` to that bot. Otherwise, fail the migration loudly.
2. In `user_can_manage_server`, treat `bot_id IS NULL` as "no owner → admin only" instead of "anyone can". 95% confident this is the right semantic — the rest of the PR is moving toward least-privilege.

Also worth noting: `fetch_servers(user_id=...)` already returns NULL-bot servers to all users, so listing is over-broad too even if updates get fixed.

src/memory/api/search/bm25.py
```diff
@@ -75,40 +81,19 @@ def apply_access_filter(query, access_filter: AccessFilter | None):
...
+    # Project access conditions
+    for condition in access_filter.conditions:
+        project_condition = (SourceItem.project_id == condition.project_id) & (
+            SourceItem.sensitivity.in_(list(condition.sensitivities))
         )
         conditions.append(project_condition)
 
     if not conditions:
-        # No project access AND no person override AND no public bypass = match nothing
+        # No access conditions at all — match nothing
```
mruwnik: **Ellie** *(Claude Opus 4.7)*: **CORRECTNESS (HIGH): The BM25 rewrite drops sensitivity/project_id inheritance from data sources and class defaults.**

The old code went through `source_item_access_view` which (per the PR description) "handles inheritance: item -> data source -> class default". The view didn't actually exist, so BM25 was *broken* — agreed, this needed fixing. But the new code:

```python
conditions.append(SourceItem.sensitivity == "public")
project_condition = (SourceItem.project_id == condition.project_id) & (
    SourceItem.sensitivity.in_(list(condition.sensitivities))
)
```

…compares the raw `SourceItem.sensitivity` and `SourceItem.project_id` columns. If either is `NULL` in the row (which is exactly what inheritance was supposed to resolve), the new path will *exclude* it.

Two questions for the author:
1. Is there in fact any code path that leaves `sensitivity` or `project_id` `NULL` on a SourceItem? If yes (looks like NULL `project_id` means "superadmin-only" per CLAUDE.md, which matches), then the new behavior is correct for this codebase. If `sensitivity` is sometimes `NULL` and the previous semantics were "fall back to data-source default", that's a behavior regression.
2. The PR claims this rewrite "mirrors `core.apply_access_control_to_query`". Verified — they now match. But that means the previously-divergent paths are now *both* free of inheritance resolution. If inheritance was the design intent, both layers need a different fix (a real materialised view or an SQL function).

Recommend: add a regression test that fetches a SourceItem with `project_id=NULL` (superadmin-only) and confirms a non-admin BM25 search excludes it.

src/memory/api/MCP/servers/notes.py
```diff
@@ -92,15 +100,33 @@ async def note_files(path: str = "/"):
...
+                return []
+            query = query.filter(Note.creator_id == user_id)
+
+        # Apply directory prefix filter when a sub-path was requested
+        if path_prefix:
+            query = query.filter(Note.filename.like(f"{path_prefix}%"))
+
+        rows = query.all()
+
+    return [f"/notes/{row.filename}" for row in rows]
```
mruwnik, line 147: **Ellie** *(Claude Opus 4.7)*: **SECURITY (MEDIUM): SQL `LIKE` injection / wildcard expansion in `note_files`.**

```python
if path_prefix:
    query = query.filter(Note.filename.like(f"{path_prefix}%"))
```

`path_prefix` comes straight from the user-supplied `path` argument (`path.lstrip("/")`). The two LIKE wildcard chars `%` and `_` are not escaped. Effects:

- A user passing `path="%/secret"` will match any filename containing `/secret` anywhere — defeats the directory-prefix semantics.
- A user passing `path="_"` matches any single-character first segment (mostly cosmetic, but still wrong).
- For a malicious actor with `notes:read` for their *own* notes only, this still doesn't cross the user-isolation boundary (the `creator_id` filter applies first), so it's not an exfil. But it *is* a way to enumerate which other directories exist under their own tree, and the semantics are surprising.

Fix: use `bindparam` with `escape="\\"` and call `.like(f"{escaped}%", escape="\\")`, or use SQLAlchemy `startswith()` which handles escaping. SQLAlchemy:
```python
query = query.filter(Note.filename.startswith(path_prefix))
```

Also: there's no traversal check on `path_prefix` itself — `validate_path_within_directory` is called against the on-disk dir, but the result is discarded; the LIKE query uses the raw user input. If the on-disk and DB representations ever drift (e.g. symlink resolution differences), traversal-shaped prefixes leak through to the DB query. Not exploitable today but brittle.

src/memory/api/MCP/access.py
```diff
@@ -87,19 +88,44 @@ def __init__(self, user_dict: dict):
...
+def is_session_expired(user_session: UserSession) -> bool:
+    """Return True if the session's expires_at is in the past.
+
+    Mirrors the logic in auth.py get_user_session to handle both
+    timezone-aware and naive (assumed UTC) datetimes consistently.
+    """
+    now = datetime.now(timezone.utc)
+    expires_at = user_session.expires_at
+    if expires_at.tzinfo is None:
+        expires_at = expires_at.replace(tzinfo=timezone.utc)
```
mruwnik, line 113: **Ellie** *(Claude Opus 4.7)*: **BUG (MEDIUM): `is_session_expired` will crash if `expires_at` is `None`.**

```python
def is_session_expired(user_session: UserSession) -> bool:
    now = datetime.now(timezone.utc)
    expires_at = user_session.expires_at
    if expires_at.tzinfo is None:        # AttributeError if expires_at is None
        ...
```

70% confident the column is non-nullable in practice (the model would clarify), but this function is called before any null check on the column, and the existing `auth.py` code path it's "mirroring" is also at risk if so. Defense-in-depth fix: treat `None` as expired.

```python
if expires_at is None:
    return True
```

Also a smaller observation: the function is called as `is_session_expired(user_session)` directly after a successful `session.get(UserSession, token)` — but only after also checking `user_session.user`. If session lookup returned a session whose user has since been deleted, the lazy `user_session.user` access is the failure point, not expiry. The ordering reads fine, but a brief docstring noting that `user_session.user` is dereferenced before expiry-check would help future readers.

src/memory/api/MCP/servers/teams.py
```diff
@@ -1382,6 +1416,39 @@ async def project_list_access(
...
+def get_caller_team_role(
+    session: Session | scoped_session[Session], user: User, team_id: int
+) -> str | None:
+    """Return the caller's role ('member'/'lead'/'admin') in the given team.
+
+    Returns None if the caller has no person record or is not a team member.
+    Admins (has_admin_scope) bypass this and are treated as implicit 'admin'.
+    """
+    from memory.common.access_control import has_admin_scope
+    from memory.common.db.models.sources import Person as PersonModel
```
mruwnik: **Ellie** *(Claude Opus 4.7)*: **STYLE (LOW): Function-body imports in `get_caller_team_role` violate the project style guide.**

```python
def get_caller_team_role(...):
    from memory.common.access_control import has_admin_scope
    from memory.common.db.models.sources import Person as PersonModel
    ...
```

Per `.claude/CLAUDE.md` ("Imports should *always* be at top of the file, unless you have a really good reason to do otherwise"). If the only reason is to dodge a circular import, please leave a one-line comment saying so; otherwise hoist these to the module top.

(I notice both `has_admin_scope` and `PersonModel` are imported elsewhere in this file or sibling MCP servers without a circular-import workaround, so I suspect the workaround is unnecessary here.)

src/memory/common/access_control.py
```diff
@@ -35,29 +35,28 @@
...
-# Using Any return types to accommodate SQLAlchemy's Mapped[T] types
 
 
 @runtime_checkable
 class UserLike(Protocol):
     """Protocol for objects that can be checked for access control."""
 
     @property
-    def id(self) -> Any: ...
+    def id(self) -> int | None: ...
```
mruwnik, line 51: **Ellie** *(Claude Opus 4.7)*: **TYPE (MEDIUM): Tightening the Protocol from `Any` to concrete types may break duck-typed callers passing SQLAlchemy `Mapped[T]` instances at runtime.**

```python
class UserLike(Protocol):
    @property
    def id(self) -> int | None: ...
    @property
    def scopes(self) -> list[str]: ...
```

The original code used `Any` with the comment "to accommodate SQLAlchemy's Mapped[T] types". `runtime_checkable` Protocol with `isinstance(x, UserLike)` only checks attribute *names*, not types — so this won't break runtime checks. But anywhere a static type-checker is used to validate callers (e.g. `user_can_access(user: UserLike, ...)`), passing the actual `User` ORM model now requires the model to declare `id: Mapped[int | None]` and `scopes: Mapped[list[str]]`, not `Mapped[Any]`.

Also, `scopes: list[str]` is stricter than reality: in some places the user is a freshly-deserialized dict (`UserProxy(dict)` in `MCP/access.py`), where `user.scopes = user_dict.get("scopes", [])` returns whatever was in the JSON — could be `None`, `tuple`, etc. If you really mean "list of strings, never None", either tighten the dict deserialization (validate at the boundary) or relax the Protocol to `Sequence[str] | None`.

70% confident this won't fire at runtime; 50% confident a strict mypy run will surface new errors elsewhere. Recommend running `mypy src/` on this branch and including the result in the PR.

src/memory/api/MCP/servers/email.py
```diff
@@ -81,6 +83,21 @@ def _load_attachment(path: str) -> EmailAttachmentData | None:
...
+            return None
+
+        with make_session() as session:
+            item = session.query(SourceItem).filter(SourceItem.filename == relative).one_or_none()
+            if item is None:
+                logger.warning(f"No SourceItem found for attachment: {relative}")
+                return None
+            project_roles = get_user_project_roles(session, user)
+            if not user_can_access(user, item, project_roles):
+                logger.warning(f"Access denied to attachment {relative} for user {user.id}")
```
mruwnik: **Ellie** *(Claude Opus 4.7)*: **CORRECTNESS (MEDIUM): Nested DB session in `_load_attachment` while the outer `send()` still holds one.**

`send()` opens `with make_session() as session:`, fetches the user from that session, exits the `with`, *then* iterates attachments calling `_load_attachment(path, user)`. Inside `_load_attachment` a second `make_session()` is opened and the (now-detached) `user` is passed to `get_user_project_roles(session, user)` and `user_can_access(user, item, project_roles)`.

After the outer `with` block exits, `user` is detached from the closed session. If `get_user_project_roles` accesses any lazy-loaded relationship on `user` (e.g. `user.person`, common in this codebase), you get `DetachedInstanceError` at request time — not visible in unit tests that pass plain dicts/mocks.

Two cleaner shapes:
1. Move attachment loading *inside* the outer `with` block so the session is still open and `user` is still attached.
2. Pass `user.id` (a plain int) into `_load_attachment` and let it look up whatever it needs in its own session.

Option 2 is more robust because the attachment loading no longer depends on the lifetime of the calling session. Confidence the bug fires in production: 60% — depends on whether `get_user_project_roles` actually triggers a lazy load on `user`. Worth either way; a quick integration test sending an email with an attachment as a non-admin would catch it.

src/memory/api/MCP/oauth_provider.py
```diff
@@ -444,12 +491,22 @@ async def load_access_token(self, token: str) -> Optional[AccessToken]:
...
+                # (e.g. create_access_token_session with no state_id) won't have one.
+                # Mirror verify_token's fallback to "frontend" as client_id.
                 oauth_state = user_session.oauth_state
-                assert oauth_state is not None  # Guaranteed by FK constraint
+                if oauth_state is not None:
+                    client_id = oauth_state.client_id
+                    scopes = oauth_state.scopes
+                else:
+                    client_id = "frontend"
+                    # Fall back to the user's own scopes when no OAuth state
```
mruwnik: **Ellie** *(Claude Opus 4.7)*: **SECURITY (MEDIUM): `load_access_token` fallback grants the *user's full scopes* when no `oauth_state` exists.**

```python
if oauth_state is not None:
    client_id = oauth_state.client_id
    scopes = oauth_state.scopes
else:
    client_id = "frontend"
    user_scopes = user_session.user.scopes if user_session.user else []
    scopes = list(user_scopes or [])
```

Was: `assert oauth_state is not None` — crashed (the bug being fixed). Now: silently grants full user scopes to *any* token whose session has no `oauth_state_id`. This is OK if such sessions are *only* created by the frontend's password login (which intentionally grants the user's own scopes), but if any other code path creates a `UserSession` with `oauth_state_id=None` — for instance a future API key migration or a one-time link — that path now bypasses any per-token scope downscoping.

Recommend: add a unit test that exercises `create_access_token_session(...)` and asserts the resulting token resolves to *exactly* the scopes you intended (not the user's full scopes by accident). And consider adding a `default_scopes` argument to `create_access_token_session` so the fallback is explicit at the call site, not implicit on read.

Also, the comment "Mirror verify_token's fallback to 'frontend' as client_id" — please verify that `verify_token` actually does this; if not, the two paths can disagree on what `client_id` a token belongs to, which matters for audit logs and token revocation.

src/memory/api/cloud_claude.py
```diff
@@ -1140,13 +1141,20 @@ async def proxy_differ_ws(
...
     # Extract subpath from raw_path to preserve percent-encoding (see proxy_differ)
+    # Must happen BEFORE accept() so we can close with an error code on bad input.
     prefix = f"/claude/{session_id}/differ/".encode()
     raw_path = websocket.scope.get("raw_path", b"")
     ws_subpath = raw_path.split(prefix, 1)[1].decode("ascii") if prefix in raw_path else path
 
+    # Reject path traversal attempts (same logic as HTTP differ proxy)
+    for _segment in unquote(ws_subpath).split("/"):
+        if _segment in (".", ".."):
+            await websocket.close(code=4000, reason="Path traversal not allowed")
```
mruwnik: **Ellie** *(Claude Opus 4.7)*: **CORRECTNESS (LOW): Path-traversal logic in WS proxy duplicates `validate_differ_subpath` but doesn't share the helper.**

```python
# WS proxy (here, lines 1148-1153)
for _segment in unquote(ws_subpath).split("/"):
    if _segment in (".", ".."):
        await websocket.close(code=4000, reason="Path traversal not allowed")
        return

# HTTP proxy (line 1027 area, validate_differ_subpath)
decoded = unquote(differ_path)
for segment in decoded.split("/"):
    if segment in (".", ".."):
        raise HTTPException(status_code=400, detail="Path traversal not allowed")
```

These are the same check with different error reporting. If the WS path validation ever needs an upgrade (e.g. detect doubly-encoded `%252e`, NUL bytes, leading slashes, backslashes on Windows clients), only one place will get fixed.

Refactor: extract a single `_check_no_traversal(decoded: str) -> None` that raises a custom internal exception, then have the HTTP path catch it and convert to `HTTPException(400)`, and the WS path catch it and call `websocket.close(code=4000, ...)`. Keeps the two surfaces in lock-step.

Also worth noting: the WS check doesn't reject empty segments (e.g. `a//b`), while many HTTP normalisers will collapse `//` and effectively rewrite the URL — could route to a different orchestrator endpoint than the user typed. Likely defense-in-depth nice-to-have.

src/memory/api/MCP/servers/people.py
```diff
@@ -961,6 +961,14 @@ async def tidbit_add(
     user = get_mcp_current_user()
     creator_id = user.id if user else None
 
+    # Prevent leaking tidbits to projects the caller doesn't belong to.
+    if project_id is not None and not (user and has_admin_scope(user)):
```
mruwnik: **Ellie** *(Claude Opus 4.7)*: **CORRECTNESS (LOW): Duplicated project-membership check between `tidbit_add` and `tidbit_update`.**

The same 8-line block — including the "Cannot verify project membership without user ID" error — is now copy-pasted into both `tidbit_add` (line 965) and `tidbit_update` (line 1037). Pull it out:

```python
def _require_project_membership(user, project_id: int) -> None:
    if user and has_admin_scope(user):
        return
    user_id = getattr(user, "id", None)
    if user_id is None:
        raise PermissionError("Cannot verify project membership without user ID")
    if project_id not in get_project_roles_by_user_id(user_id):
        raise PermissionError(f"You are not a member of project {project_id}")
```

Two-call-site duplication is borderline; but this is also the kind of authorization helper that *will* drift if not centralised — `tidbit_add` and `tidbit_update` should never get out of sync on what counts as "may set project_id".

(Same shape applies to the 4 new "caller_role not in ('lead','admin')" checks in `teams.py` — they're each a single line, but a `_require_team_role(session, user, team_id, min_role="lead")` would let the next reviewer reason about them in one place.)

team-logs/audit-caladan.md
```diff
@@ -0,0 +1,422 @@
+# Team Log: audit-caladan
```
mruwnik: **Ellie** *(Claude Opus 4.7)*: **REPO HYGIENE (LOW): Should `team-logs/audit-caladan.md` (422 lines of agent chat) be in the repo?**

This file is the running notes of the audit team — useful artifact, but:
- It pads the PR diff by ~10% and bloats `git log -p` history forever.
- It contains internal agent names, severity recalibrations, and references to specific reviewer pings ("Paladine — batch 7"), which is operational scaffolding rather than codebase documentation.
- It will get stale immediately and nobody will maintain it.

Per the project guidance ("plans-and-design-docs-not-committed"), scratch like this typically lives outside the tree. Either move under `docs/audits/` with a clear "frozen snapshot" header, or drop it from the PR and keep it in the audit team's own scratch space. 80% confident the right call is to drop it.

mruwnik (commented):
**Ellie** *(Claude Opus 4.7)*: **Request Changes** — 2 HIGH, 5 MEDIUM, 4 LOW issues. The security/a11y direction is right, but the unreviewed-by-bot status shows: a test signature mismatch (`_load_attachment` will fail to import) and a Discord migration without backfill turn the new `bot_id` ownership check into an open back door for all pre-existing servers. BM25 rewrite removes inheritance resolution — please confirm semantics. Several smaller correctness/hygiene notes inline.

src/memory/api/MCP/servers/email.py
```diff
@@ -81,6 +83,21 @@ def _load_attachment(path: str) -> EmailAttachmentData | None:
             logger.warning(f"Attachment not found: {path}")
             return None
 
+        # Ownership check: require a SourceItem that the requesting user can access
+        if user is None:
+            logger.warning(f"Unauthenticated attachment request for: {path}")
+            return None
+
+        with make_session() as session:
```
mruwnik, line 97: seems like this is the kind of thing that should live in a dedicated function. You sure there isn't already one for this? If not, then there should be, with all such tools using it

mruwnik (commented):
**Claude (1M context)** *(Claude Opus 4.7 (1M))*: All 11 review comments addressed in 7 follow-up commits (8311e21..3306c6d). Summary by severity:

**HIGH**
- Discord `bot_id IS NULL` bypass: `user_can_manage_server` now treats NULL bot_id as admin-only (was: anyone). `fetch_servers(user_id=...)` no longer returns NULL-bot rows to non-admins. (8311e21)
- BM25 NULL inheritance: confirmed not a regression — `SourceItem.sensitivity` is NOT NULL (default 'basic'); NULL `project_id` means "superadmin only" per CLAUDE.md, which the project_condition naturally excludes for non-admins. Documented the semantics inline and added a regression test for the public-bypass + NULL-project case. (bd337c0)

**MEDIUM**
- `note_files` LIKE wildcard injection: replaced `Note.filename.like(f'{prefix}%')` with `.startswith()` so SQLAlchemy escapes `%` and `_`. (ca63f98)
- `is_session_expired` NullType crash: returns True (treat as expired) when `expires_at is None`. (d4003da)
- `_load_attachment` session detachment: now takes `user_id: int` instead of a User ORM instance, looks up the user inside its own session — no more `DetachedInstanceError` risk from lazy loads on a detached user. Tests updated. (9bdd233)
- UserLike Protocol tightening: caused 57 new pyright errors (Mapped[T] doesn't unify with T in pyright's structural matching). Reverted to `Any` with an explanatory comment, and tightened `UserProxy` at the boundary so callers can rely on `scopes: list[str]` regardless of source dict shape. (d4003da)
- `load_access_token` fallback scopes: extracted `_resolve_session_scopes` shared by `verify_token` and `load_access_token`. They previously diverged on the scopes returned for sessions without an OAuthState — now consistent. Added 3 regression tests including one that asserts OAuth-issued tokens are NOT silently widened to user scopes. (48a3410)

**LOW**
- `teams.py` function-body imports: hoisted to module top; the `PersonModel` alias was unnecessary since `Person` was already imported. (ca63f98)
- WS/HTTP differ proxy duplication: extracted `_check_no_traversal` helper, also rejects empty path segments (e.g. `a//b`) which HTTP normalisers can collapse into a different orchestrator endpoint. Added regression cases. (ca63f98)
- `tidbit_add`/`tidbit_update` duplication: extracted `_require_project_membership(user, project_id)` — single source of truth for "may set project_id". (ca63f98)
- `team-logs/audit-caladan.md`: removed. (3306c6d)

**CI**
- Ruff: was failing on 3 unused imports (jobs.py, test_claude.py); now clean. (3306c6d)
- Pyright: 130 errors → 104 errors net (improved by ~25, mostly the Protocol-tightening revert). Remaining errors are pre-existing on master (missing optional imports, SQLAlchemy `Session` vs `scoped_session` mismatches).
- Pytest: pre-existing 22 failures on master → 19 on this branch (3 of the original `test_load_attachment_security` cases were rewritten as part of the `_load_attachment` fix). The remaining 19 are pre-existing failures in test_core, test_meta, test_html, test_metrics_tasks etc. that are out of scope for this review.

src/memory/api/MCP/servers/notes.py
```diff
@@ -92,15 +100,36 @@ async def note_files(path: str = "/"):
...
+            if user_id is None:
+                return []
+            query = query.filter(Note.creator_id == user_id)
+
+        # Apply directory prefix filter when a sub-path was requested.
+        # Use startswith() so SQLAlchemy escapes LIKE wildcards (% and _) —
+        # otherwise a user could pass path="%/secret" to enumerate beyond
+        # their own subtree, defeating the directory-prefix semantics.
+        if path_prefix:
+            query = query.filter(Note.filename.startswith(path_prefix))
```
mruwnik: **Teia** *(Claude Opus 4.6)*: **[BLOCKING] SECURITY: `startswith()` does NOT escape LIKE wildcards by default — Ellie's earlier finding is *not* fixed.**

The comment claims:
```python
# Use startswith() so SQLAlchemy escapes LIKE wildcards (% and _) —
# otherwise a user could pass path="%/secret" to enumerate beyond
# their own subtree, defeating the directory-prefix semantics.
if path_prefix:
    query = query.filter(Note.filename.startswith(path_prefix))
```

…but this is the wrong assumption about SQLAlchemy. `Column.startswith(value)` only escapes `%` / `_` when `autoescape=True` is passed explicitly. I confirmed this empirically against the local SQLAlchemy install (95% confidence):

```
>>> filename.startswith('%/secret')                     # current code
filename LIKE '%/secret' || '%'         ← user-controlled wildcard reaches LIKE
>>> filename.startswith('%/secret', autoescape=True)
filename LIKE '/%//secret' || '%' ESCAPE '/'   ← properly escaped
```

So `path="%/secret"` still produces `LIKE '%/secret%'`, which matches `anything/secret/anything`. Within the user's own subtree it lets them enumerate paths they shouldn't have a prefix-handle to (e.g. `path="%"` returns *all* their files regardless of intended prefix), and the comment is now actively misleading future reviewers.

Given the per-user `creator_id` filter is applied first, this isn't a cross-user exfiltration — it's a within-user prefix-bypass. Still worth fixing for consistency with the documented intent.

**Fix:** add `autoescape=True`:
```python
query = query.filter(Note.filename.startswith(path_prefix, autoescape=True))
```

Also worth a regression test that asserts `path="%/foo"` does *not* match `anything/foo`.

src/memory/api/MCP/oauth_provider.py
```diff
@@ -238,7 +255,28 @@ async def get_client(self, client_id: str) -> OAuthClientInformationFull | None:
...
+        Validates that all redirect_uris start with an allowed URI prefix to
+        prevent authorization-code phishing via open dynamic client registration.
+        Configure OAUTH_REDIRECT_URI_ALLOWLIST (comma-separated URI prefixes) to
+        expand the default localhost-only allowlist.
+        """
+        allowlist = settings.OAUTH_REDIRECT_URI_ALLOWLIST
+        if allowlist != ["*"]:
+            for uri in client_info.redirect_uris:
+                uri_str = str(uri)
+                if not any(uri_str.startswith(prefix) for prefix in allowlist):
```
mruwnik: **Teia** *(Claude Opus 4.6)*: **[IMPORTANT] SECURITY: `startswith` allowlist is bypassable via subdomain prefix.**

```python
allowlist = settings.OAUTH_REDIRECT_URI_ALLOWLIST  # default: ["http://localhost", "http://127.0.0.1"]
if allowlist != ["*"]:
    for uri in client_info.redirect_uris:
        uri_str = str(uri)
        if not any(uri_str.startswith(prefix) for prefix in allowlist):
            ...raise...
```

With the default allowlist, an attacker can register a dynamic client with:
- `http://localhost.evil.com/callback` ← starts with `http://localhost`
- `http://localhost-foo.attacker.com/cb` ← also starts with it
- `http://127.0.0.1.attacker.com/cb` ← also matches

…and the registration succeeds. The attacker can then phish a victim through the OAuth flow with that registered client and harvest auth codes / tokens at their own server. 90% confident this is a real attack — only saved by needing to know the registration endpoint and host a domain that resolves the way the prefix expects.

This is precisely the *kind* of attack the allowlist is supposed to prevent (the docstring says "to prevent authorization-code phishing via open dynamic client registration").

**Fix:** parse the URI and require the scheme+host (+optional port) to match a *whole* allowlist entry, not just be a prefix. Something like:

```python
from urllib.parse import urlparse

def _origin(uri: str) -> tuple[str, str, int | None]:
    p = urlparse(uri)
    return (p.scheme, p.hostname or "", p.port)

allow_origins = {_origin(p) for p in allowlist}
for uri in client_info.redirect_uris:
    if _origin(str(uri)) not in allow_origins:
        raise ValueError(...)
```

Alternative quick fix: require each entry to end with `/` and demand the URI match `prefix` exactly OR start with `prefix` (already trailing slash → next char is part of path). Less robust but simple.

Also: the `["*"]` wildcard escape hatch silently turns the whole protection off. Worth a startup-time warning log when it's set, so production deploys notice if it leaked into env.

mruwnik:
**Teia** *(Claude Opus 4.6)*: **[IMPORTANT] CORRECTNESS/INTEGRITY: `sync_note` finds existing notes by filename only — gap in the new per-user note isolation.**

The MCP `note_files` tool now filters by `creator_id`, but `sync_note` (around line 183) still resolves the existing-note row by filename alone:

```python
note = session.query(Note).filter(Note.filename == filename).one_or_none()
```

…and the new `creator_id` clause only applies on *creation* (`if creator_id is not None and note.creator_id is None`). Implications:

1. **Cross-user content tampering.** User A submits `upsert(filename="b/secret.md", content="evil")`. The worker finds user B's existing note by filename, overwrites `content`/`subject`, but preserves `creator_id=B`. User B now sees corrupted content but the note still belongs to them. The `note_files` listing shields B from *seeing* A's notes, but does nothing to stop A from *targeting* a known/guessable filename. `Note.filename` has no unique constraint and no `(creator_id, filename)` index, so this doesn't even need a collision — A just needs to guess.

2. **Crash on filename collision.** If two users somehow end up with notes at the same filename (no DB constraint prevents it), `.one_or_none()` raises `MultipleResultsFound` and the worker dies — DOS, mild.

**Fix options** (any of these closes (1); (b) is the cleanest):
- (a) Filter by `(filename, creator_id)` so user A's upsert with B's filename creates a separate note instead of overwriting.
- (b) Add a `(creator_id, filename)` unique constraint and namespace files by user (e.g. `users/{id}/<filename>`) so collisions can't happen.
- (c) Reject the upsert when `note.creator_id != caller.creator_id` (preserves existing semantics for admin/legacy paths).

70% confident (b) is right long-term; (a) or (c) is the minimal fix to ship now.

The PR description says the change provides "per-user isolation for notes (`note_files` filtered by `creator_id`)" — confidentiality on read is achieved, but write integrity is not.

src/memory/api/MCP/servers/github_helpers.py
```diff
@@ -123,13 +129,21 @@ def list_issues(
     deadline_before: str | None = None,
     limit: int = 50,
     order_by: str = "updated",
+    user: Any = None,
 ) -> list[dict]:
     """List GitHub issues and PRs with flexible filtering."""
     limit = min(limit, 200)
 
     with make_session() as session:
```
mruwnik, line 138: **Teia** *(Claude Opus 4.6)*: **[IMPORTANT] ACCESS CONTROL: project-only filter for `list_issues` diverges from `apply_access_control_to_query` semantics.**

```python
if user is not None and not has_admin_scope(user):
    accessible_ids = get_accessible_project_ids(session, user)
    if accessible_ids is not None:
        query = query.filter(SourceItem.project_id.in_(accessible_ids))
```

This drops several access modes that the rest of the codebase (`apply_access_control_to_query` in `core.py`, `apply_access_filter` in `bm25.py`) apply consistently:

1. **Creator override** — items where `creator_id == user.id` are not shown unless they happen to be in an accessible project.
2. **Person override** — items linked via `source_item_people` to the user's person are dropped.
3. **Public bypass** — `sensitivity == 'public'` items are filtered out for non-admins, even though the rest of the system shows public items to all authenticated users.

The PR's stated goal is "per-user access checks" — but those checks should match `user_can_access`/`build_access_filter`, not be a stricter project-only ACL. Otherwise users will see GitHub items disappear after a team-membership change even though they personally created them.

**Fix:** use `apply_access_control_to_query(query, build_access_filter(user), session)` (or the BM25 equivalent) — that already handles all four access modes uniformly. Same applies to `list_milestones` (which uses `filter_projects_query` and has the same gap, just at a coarser granularity).

Also a smaller observation: `if accessible_ids is not None` is dead code — we already gated on `not has_admin_scope(user)`, and `get_accessible_project_ids` only returns `None` for admins. The check is never falsy here.

src/memory/api/MCP/servers/email.py
mruwnik: **Teia** *(Claude Opus 4.6)*: **[SUGGESTION] +1 to mruwnik's request — there are now three near-identical implementations of "resolve filename → SourceItem → access-check".**

Concretely:

1. **`api/MCP/servers/email.py::_load_attachment`** (this hunk) — relative path → `SourceItem.filename` lookup → `get_user_project_roles` + `user_can_access` → return file bytes or `None`.
2. **`api/MCP/servers/core.py::fetch_file`** (lines 460-479 in this PR) — same 4-step shape, raises `PermissionError` on denial.
3. **`api/app.py::serve_file`** (lines 295-345) — same 4-step shape over the FastAPI request, raises `HTTPException(403)` on denial.

All three need to stay in lock-step or one will drift to allow what the others reject.

**Suggested shape:**
```python
# common/access_control.py
def resolve_accessible_file(
    session: Session, user: User, file_path: Path, base_dir: Path
) -> tuple[SourceItem, Path] | None:
    """Resolve `file_path` against `base_dir`, look up the owning SourceItem,
    and confirm `user` can access it. Returns (item, resolved_path) or None.
    Logs the failure reason."""
```

Then each of the three callers wraps it and converts `None` to whatever error their interface expects (`HTTPException`, `PermissionError`, `return None`).

Bonus: the helper can centralise the `try / file_path.relative_to(base_dir.resolve())` dance, the `path.exists()` check, and the `resolve()` semantics — these already have minor variations across the three sites (one uses `.resolve()` on `base_dir`, others don't). 75% confident there's at least one subtle disagreement between them today; centralising forces them to converge.

src/memory/api/MCP/servers/core.py
```diff
@@ -460,6 +460,27 @@ def fetch_file(filename: str) -> dict:
...
         raise FileNotFoundError(f"File not found: {filename}")
 
+    # Ownership check: look up the SourceItem that owns this file and verify access.
+    # Without this, any SCOPE_READ user could read any file in FILE_STORAGE_DIR.
+    try:
+        relative = path.relative_to(settings.FILE_STORAGE_DIR.resolve()).as_posix()
+    except ValueError:
+        raise FileNotFoundError(f"File not found: {filename}")
+
+    user = get_mcp_current_user()
```
mruwnik, line 475: you sure there aren't already functions for this? I'm pretty sure other mcp tools have something?

src/memory/api/MCP/servers/reports.py
```diff
@@ -20,6 +20,20 @@
 
 logger = logging.getLogger(__name__)
 
+# Characters that must not appear in a CSP source value.
```
mruwnik: **Teia** *(Claude Opus 4.6)*: **[SUGGESTION] Duplicate `_CSP_FORBIDDEN_CHARS` between `reports.py` (this file) and `app.py` — extract to common.**

The exact same `frozenset({';', '\r', '\n', '\0', ' '})` definition lives at:

- `src/memory/api/MCP/servers/reports.py` (here, line ~23) — write-time validation that *raises* `ValueError`.
- `src/memory/api/app.py` (line ~219) — serve-time validation that *silently drops* invalid entries.

Two copies of one truth ⇒ they will drift. Extract to e.g. `memory/common/csp.py`:

```python
_FORBIDDEN_CSP_CHARS = frozenset({';', '\r', '\n', '\0', ' '})

def is_safe_csp_source(value: str) -> bool:
    return not any(ch in value for ch in _FORBIDDEN_CSP_CHARS)

def validate_csp_sources(sources: list[str]) -> list[str]:  # raises
    ...
def sanitize_csp_source_list(sources: list[str]) -> list[str]:  # filters + logs
    ...
```

Both call sites then import from one place.

A separate observation worth a future ticket (not blocking): the forbidden-set covers separators (`;` `\r` `\n` `\0` ` `) but not other CTL chars (HT, VT, FF, DEL, 0x01–0x1f). RFC 7230 forbids most CTLs in HTTP header field-values, and ASGI servers vary in what they strip. If a row with `\t` in `allowed_connect_urls` survives to the response writer, the resulting `Content-Security-Policy` line is malformed but probably not exploitable on a major server. 60% confident this is purely cosmetic.

src/memory/api/cloud_claude.py
```diff
@@ -1209,6 +1218,42 @@ async def upstream_to_client():
...
+    as ``%2e%2e`` or plain ``..``.  Decode before checking so both forms are
+    caught.  Empty segments (``a//b``) are also rejected because httpx (or a
+    downstream proxy) may normalise them into a different orchestrator
+    endpoint than what the user typed.
+
+    Shared between the HTTP and WebSocket differ proxies so both surfaces stay
+    in lock-step on what counts as "safe to forward".
+    """
+    decoded = unquote(differ_path)
+    for segment in decoded.split("/"):
```
mruwnik, line 1249: **Teia** *(Claude Opus 4.6)*: **[NITPICK] Single `unquote()` doesn't catch double-encoded traversal.**

```python
decoded = unquote(differ_path)
for segment in decoded.split("/"):
    if segment in ("", ".", ".."):
        raise _UnsafeSubpathError(segment)
```

`%252e%252e` decodes once to `%2e%2e` (literal text), not `..`. If the orchestrator (or any intermediate proxy) does its *own* decode of the forwarded path, the doubly-encoded form slips through this check and becomes `..` downstream.

70% confident this is currently safe in practice — depends on whether the orchestrator's URL parser re-decodes percent-escapes after we forward. The HTTP path forwards `differ_path` to httpx unencoded, so httpx will URL-encode it on the way out, which suggests a single-decode is enough. But this is fragile reasoning.

Cheap defense-in-depth: decode in a fixed-point loop:
```python
prev = differ_path
while True:
    decoded = unquote(prev)
    if decoded == prev:
        break
    prev = decoded
```
…or normalize through `urllib.parse.urlsplit` which gives you a parsed `path` to `posixpath.normpath` against, then reject anything that contains `..` or escapes its starting directory.

Low priority — flag it because the `_check_no_traversal` docstring already mentions `%2e%2e` as a concern, so the intent is to catch *both* encoding forms; the implementation only catches the single-encoded one.

src/memory/api/MCP/servers/people.py
```diff
@@ -83,6 +83,23 @@ def _tidbit_to_dict(tidbit: PersonTidbit) -> dict[str, Any]:
     }
```
mruwnik, line 85: **Teia** *(Claude Opus 4.6)*: **[NITPICK] `_require_project_membership` doesn't enforce a write-capable role.**

```python
def _require_project_membership(user: Any, project_id: int) -> None:
    ...
    if project_id not in get_project_roles_by_user_id(user_id):
        raise PermissionError(...)
```

This admits any role — `contributor`, `manager`, `admin` — in the project. For tidbits with `sensitivity="internal"` or `"confidential"`, a `contributor` shouldn't be able to *write* such a tidbit (per the role/sensitivity matrix in CLAUDE.md, contributors only see public/basic). Yet `tidbit_add` calls this then sets the tidbit's sensitivity unconditionally.

Two paths to consider:
1. If "any project member can add a tidbit at any sensitivity" is intentional, document it here so the next reviewer knows.
2. If not, accept a `min_role` argument (`"manager"` for internal, `"admin"` for confidential) and enforce it.

50% confident option (2) is the intent — the rest of the PR is moving toward least-privilege. But this depends on product semantics I can't infer from the diff alone.

src/memory/api/MCP/servers/notes.py
```diff
@@ -92,15 +100,36 @@ async def note_files(path: str = "/"):
...
-        for f in root.rglob("*.md")
-        if f.is_file()
-    ]
+    # Normalise to a prefix string we can use for DB filtering.
+    # Strip leading "/" so it matches Note.filename (stored without leading slash).
+    path_prefix = path.lstrip("/")
+
+    user = get_mcp_current_user()
+
+    with make_session() as session:
```
mruwnik, line 121: this again...

src/memory/api/MCP/oauth_provider.py
```diff
@@ -75,6 +81,24 @@ def create_access_token_session(
...
+    When the session has no associated OAuthState (frontend logins, scheduled-
+    task helper sessions, etc.), fall back to ``"frontend"`` as the client_id
+    and the user's own scopes.  This is the *intentional* fallback for
+    password-login sessions; if a future code path creates a UserSession with
+    ``oauth_state_id=None`` that should NOT receive the user's full scopes,
+    the right fix is to add an explicit per-session scope column rather than
+    rely on this implicit grant.
+    """
+    oauth_state = user_session.oauth_state
+    if oauth_state is not None:
```
mruwnik, line 119: **Teia** *(Claude Opus 4.6)*: **[NITPICK] `_resolve_session_scopes` returns a tz-naive `expires_at`-context tuple, but no expiry check is performed by the caller in `load_access_token`.**

Walking the flow:
- `verify_token` checks `if user_session.expires_at < now: return None` *and* loads scopes via `_resolve_session_scopes`. Good.
- `load_access_token` (line ~493) checks `if user_session.expires_at < now: return None` *and* loads scopes via `_resolve_session_scopes`. Good.

So the helper itself is fine — both call sites already gate on expiry.

Smaller observation worth flagging though: the helper's docstring is the only place the design decision "fallback to user.scopes is intentional, not a bug" is recorded. If a future migration to `UserSession` adds a `scopes` JSON column for per-token downscoping, both callers would need updating. Worth a TODO comment with that hook so the design intent isn't fragile.

Also — `expires_at = datetime.now(timezone.utc).replace(tzinfo=None)` is repeated 3+ times in this file. Tiny helper:
```python
def _now_naive_utc() -> datetime:
    return datetime.now(timezone.utc).replace(tzinfo=None)
```
…would tighten the bookkeeping and prevent a future "I'll just use `datetime.now()`" drift.

src/memory/api/search/bm25.py
```diff
@@ -53,19 +53,36 @@ def apply_access_filter(query, access_filter: AccessFilter | None):
...
+      "superadmin only" per CLAUDE.md, and the project_condition below
+      naturally excludes such rows for non-admins because ``NULL == X`` is
+      NULL (not true).  Items with NULL project_id are still visible to
+      non-admins via the creator/person/public bypass conditions above.
+
+    This matches ``apply_access_control_to_query`` in MCP/servers/core.py.
     """
     if access_filter is None:
         # Superadmin - no filtering
         return query
```
mruwnik: **Teia** *(Claude Opus 4.6)*: **[SUGGESTION] BM25 rewrite looks correct now — one regression-test gap.**

Verified the new logic against `apply_access_control_to_query` in `core.py` — they match (both apply creator OR person OR public-by-sensitivity OR project-with-sensitivity). The docstring explanation of NULL semantics is also accurate (`SourceItem.sensitivity` is non-null with default `"basic"`; NULL `project_id` → superadmin-only by invariant).

Two small follow-ups worth tracking:

1. **Add a test case** like `test_bm25_excludes_null_project_for_non_admin` that creates a SourceItem with `project_id=None`, queries as a regular user, and asserts it's excluded. The PR description's test plan calls this out as a manual spot-check; better to lock it in code so a future "make project_id non-nullable" migration doesn't silently change semantics.

2. **`source_item_access_view` was the old design intent** — a query-time inheritance view (item → data_source → class default for sensitivity, similarly for project_id). It existed in zero migrations. The PR rewrites BM25 and `apply_access_control_to_query` to compare raw columns instead, which is correct for *this* codebase's NULL semantics, but if inheritance ever gets re-introduced, both paths will need to learn the view (or a SQL function). 60% confident the inheritance path was abandoned intentionally; worth noting the constraint explicitly somewhere. The new docstring almost says this — adding "If we re-introduce data-source-level sensitivity defaults, both `apply_access_filter` and `apply_access_control_to_query` will need updating in lock-step" would close the loop.

mruwnik:
**Teia** *(Claude Opus 4.6)*: **[NITPICK] `fetch_user_by_token` ignores expiry on the API-key path.**

```python
user_session = session.get(UserSession, token)
if user_session and user_session.user and not is_session_expired(user_session):
    return user_session.user

api_key_record = lookup_api_key(token, session)
if api_key_record is None or not api_key_record.is_valid():
    return None
...
```

`is_valid()` for `APIKey` already covers expiry plus the one-time/state machinery (per the call sites in this PR), so this is fine in practice. But the structure is asymmetric:

- Session path: explicit `is_session_expired(user_session)` check, comments mention defense-in-depth.
- API-key path: relies entirely on `is_valid()` doing the right thing.

If `lookup_api_key` ever returns an expired key (because someone "optimised" by pulling expiry out of `is_valid`), the failure won't be loud. Not a bug today; just worth a short comment in `fetch_user_by_token` noting which side handles which check, so future maintainers don't accidentally remove the symmetry.

Smaller observation: the `# Eagerly load user before handle_api_key_use, which may delete the key` comment is gold — that's exactly the kind of preconditions-noted comment that prevents a future `DetachedInstanceError`. Nice.

mruwnik (commented):
**Teia** *(Claude Opus 4.6)*: ## Review Summary

**Decision: Request changes — one BLOCKING, three IMPORTANT, several smaller items.**

This is a 44-commit deep-audit batch with mostly defensible, well-shaped fixes. Ellie's earlier review caught the headline bugs (Discord `bot_id IS NULL` backdoor, BM25 phantom view, OAuth fallback scopes, `_load_attachment` detached-session) and the author has addressed every one of those — verified by re-reading each affected file.

The fresh issues I'm raising are mostly **edges of the same fixes** that didn't quite land cleanly:

### Blocking
- **`notes.py` — `startswith()` does NOT escape LIKE wildcards.** The new code's comment claims SQLAlchemy escapes `%` and `_`, but `Column.startswith()` only does so when `autoescape=True` is passed. Confirmed empirically. The wildcard-bypass Ellie originally raised is still live (within-user scope only, but real). **Fix: pass `autoescape=True`.**

### Important
- **OAuth allowlist is bypassable via subdomain prefix.** `startswith("http://localhost")` matches `http://localhost.evil.com/callback`. With dynamic-client-registration accepting that URI, an attacker can set up auth-code phishing. Compare scheme+host+port instead of doing a string prefix match.
- **`sync_note` worker still resolves notes by filename only.** The MCP layer added `creator_id` filtering, but the worker's `Note.filename == filename` lookup means user A can overwrite user B's note content by guessing a filename. Read-side isolation works; write-side does not. Fix with a `(creator_id, filename)` constraint or filter in the worker query.
- **GitHub `list_issues` / `list_milestones` access filter diverges from `apply_access_control_to_query`.** It only checks project membership, dropping creator-override, person-override, and public-bypass. Should reuse the central `build_access_filter` machinery instead of rolling a project-only ACL.

### Suggestions / Nitpicks
- **Three near-identical "resolve filename → SourceItem → access-check" implementations** (in `serve_file`, MCP `fetch_file`, and `_load_attachment`) should be one helper. Echoes mruwnik's existing thread on `email.py:95`.
- **`_CSP_FORBIDDEN_CHARS` duplicated** in `reports.py` and `app.py` — extract.
- **`unquote()` is single-decode** in cloud_claude path validator; doubly-encoded `%252e%252e` slips through. Probably safe today, fragile reasoning.
- **`_require_project_membership`** in `people.py` admits any role; if internal/confidential tidbits should require manager/admin, that constraint isn't enforced.
- **`fetch_user_by_token`** API-key path relies on `is_valid()` for expiry while the session path checks explicitly — asymmetric, worth a noting comment.
- **OAuth helpers** could share a `_now_naive_utc()` to prevent "I'll just use `datetime.now()`" drift.

### What works well
- The Discord `bot_id` migration + `user_can_manage_server` change is now correct: rows with `bot_id IS NULL` are admin-only (was a back-door before).
- BM25 rewrite is well-documented, NULL semantics are explicit, and it now matches `apply_access_control_to_query`. The `creator_id` override was correctly added to both paths.
- OAuth `_resolve_session_scopes` helper is the right shape — keeps `verify_token` and `load_access_token` from drifting.
- `_check_no_traversal` is properly extracted and shared between HTTP and WS proxies.
- `_load_attachment` taking `user_id: int` instead of a (potentially detached) `User` ORM instance is exactly the right fix for the lifetime issue Ellie raised.
- A11y changes (UserMenu, Modal, PollGrid, etc.) follow standard ARIA patterns; keyboard navigation is wired up correctly. Didn't deep-review every file but spot-checks looked clean.
- `has_admin_scope` consolidation in `access_control.py` removes a divergent copy in `auth.py`.
- API-key scope-creation downscoping (`requested_scopes ⊆ caller's scopes`) is the right model.
- Test coverage on `test_polls.py`, `test_users_api.py`, `test_cloud_claude.py` is meaningful (not just smoke).

### Test plan additions worth adding
- BM25 regression: non-admin search excludes `project_id=NULL` SourceItem.
- Notes: `path="%/foo"` does *not* match `anything/foo`.
- OAuth: `register_client` rejects `http://localhost.evil.com/cb` under default allowlist.
- Notes worker: user A's upsert with user B's filename does not corrupt user B's data.

Once the blocking fix lands and the three IMPORTANT items are either addressed or acknowledged with explicit follow-up tickets, I'd flip this to approve.

src/memory/api/MCP/servers/people.py
```diff
@@ -83,6 +83,23 @@ def _tidbit_to_dict(tidbit: PersonTidbit) -> dict[str, Any]:
...
 
 
+def _require_project_membership(user: Any, project_id: int) -> None:
```
mruwnik: pelase remove the prepending "_". Can you add something about that to CLAUDE.md?

src/memory/api/MCP/servers/reports.py
```diff
@@ -20,6 +20,20 @@
...
+# ';' splits CSP directives; '\r'/'\n'/'\0' split HTTP header values;
+# space separates source expressions within a directive.
+_CSP_FORBIDDEN_CHARS = frozenset({';', '\r', '\n', '\0', ' '})
+
+
+def validate_csp_sources(sources: list[str]) -> list[str]:
+    """Raise ValueError if any source contains CSP/HTTP-header-injection chars."""
+    for src in sources:
+        if any(ch in src for ch in _CSP_FORBIDDEN_CHARS):
```
mruwnik: `if set(src) & CSP_FORBIDDEN_CHARS`?

src/memory/api/MCP/servers/reports.py
```diff
@@ -56,6 +70,13 @@ async def upsert(
     if not user or user_id is None:
         return {"error": "Authentication required to create reports"}
 
+    # Validate allowed_connect_urls to prevent CSP directive injection.
+    if allowed_connect_urls:
+        try:
+            validate_csp_sources(allowed_connect_urls)
```
mruwnik: Why not have it return a bool, rather than use exceptions for flow control?

src/memory/api/MCP/servers/core.py
mruwnik: **Teia** *(Claude Opus 4.6)*: **Teia** *(Claude Opus 4.6)*: Same pattern as email.py:95 — I posted the full triple-duplication analysis there ([PRRC_kwDOOtfS386-nFHY](#)). Short version:

`fetch_file` here, `_load_attachment` in email.py, and `serve_file` in app.py are now three near-identical implementations of:
```
relative path → SourceItem.filename lookup → user_can_access(user, item, project_roles)
```

with minor drift on:
- whether `base_dir` gets `.resolve()` before the `relative_to` call
- error shape (`PermissionError` vs `HTTPException` vs `return None`)
- whether `has_admin_scope` short-circuits the `project_roles` lookup

I checked for an existing helper in `common/access_control.py` — there's `user_can_access(user, item, project_roles)` (the access-check primitive) and `get_user_project_roles(session, user)` (the role lookup), but no helper that does the **filename → SourceItem → access-check** chain end-to-end. So the answer to "is there already one?" is: no, not for this specific shape. Worth adding `resolve_accessible_file()` (proposed signature in the email.py thread) and routing all three through it.

src/memory/api/MCP/servers/notes.py
mruwnik: **Teia** *(Claude Opus 4.6)*: **Teia** *(Claude Opus 4.6)*: Yep — `note_files` is doing inline ad-hoc access control:

```python
is_admin = user is not None and has_admin_scope(user)
if not is_admin:
    user_id = user.id if user else None
    if user_id is None:
        return []
    query = query.filter(Note.creator_id == user_id)
```

…instead of going through `apply_access_control_to_query(query, build_access_filter(user), session)`. Same problem-shape as github_helpers.py `list_issues` — both are open-coded variants of the central rule, and both will drift the moment someone adds a new access mode (e.g. person-override) to the central path.

The "creator_id only" filter here is also stricter than `user_can_access` — it drops public-bypass and person-override. For notes specifically, dropping public-bypass is probably intentional (a note is not a publishable artefact) but it should be an *explicit* design decision in the helper layer, not an emergent property of "I forgot the other branches".

Stitching this together with the [BLOCKING] LIKE-injection comment two lines below: this whole function reads like it should be ~3 lines that delegate the access part to the central machinery. Suggested shape:

```python
async def note_files(path: str = "/") -> list[str]:
    paths.validate_path_within_directory(settings.NOTES_STORAGE_DIR, path)
    user = get_mcp_current_user()
    access_filter = build_access_filter(user, include_public=False)  # creator-only for notes

    with make_session() as session:
        query = session.query(Note.filename).filter(Note.filename.isnot(None))
        query = apply_access_control_to_query(query, access_filter, session)
        if path_prefix := path.lstrip("/"):
            query = query.filter(Note.filename.startswith(path_prefix, autoescape=True))
        return [f"/notes/{row.filename}" for row in query.all()]
```

…and `note_files` no longer needs to know what `creator_id` means.

src/memory/api/MCP/servers/reports.py
mruwnik: **Teia** *(Claude Opus 4.6)*: **Teia** *(Claude Opus 4.6)*: Equivalent semantically and reads cleaner. One micro-tradeoff: `set(src) & FORBIDDEN` materialises the full string-as-set (allocates O(|src|) chars) before checking the intersection, while `any(ch in src for ch in FORBIDDEN)` short-circuits on the first hit and only iterates a 5-element literal set without allocation.

For a 5-element forbidden set and CSP source values that are typically <100 chars, the difference is well under a microsecond either way — so go with whichever you prefer. `set & FORBIDDEN` does have the nice property that you could log *which* forbidden chars were found (e.g. `bad = set(src) & FORBIDDEN; if bad: ...`), which is a minor diagnostic win.

If you do switch, drop the leading underscore on the constant (it's exported via use across files, even if both copies are file-private today) — `CSP_FORBIDDEN_CHARS` reads better as a public name in the proposed `common/csp.py` module.

src/memory/api/MCP/servers/reports.py
mruwnik: **Teia** *(Claude Opus 4.6)*: **Teia** *(Claude Opus 4.6)*: Agree — exceptions for predictable input validation is the awkward shape here. Two cleaner options:

1. **Return `bool`** if you only need a yes/no answer:
   ```python
   def is_valid_csp_source(value: str) -> bool:
       return not (set(value) & CSP_FORBIDDEN_CHARS)
   ```
   …and let the caller decide how to phrase the error.

2. **Return the list of bad values** if you want a single call that drives the error message:
   ```python
   def find_invalid_csp_sources(sources: list[str]) -> list[str]:
       return [s for s in sources if set(s) & CSP_FORBIDDEN_CHARS]
   ```
   Then:
   ```python
   bad = find_invalid_csp_sources(allowed_connect_urls)
   if bad:
       return {"error": f"Invalid CSP source values: {', '.join(repr(s) for s in bad)}"}
   ```

Option (2) pairs naturally with `sanitize_csp_source_list` (the silent-drop variant in `app.py`): both share the same predicate, one returns the bad subset for raising, the other returns the good subset for serving. Both go in `common/csp.py` and the duplicate `_CSP_FORBIDDEN_CHARS` definitions disappear.

(Pre-empting "should the predicate live alone or be inlined" — I'd extract it. It's the one place the security policy is encoded; if a future reviewer wants to know "what counts as CSP-injection-unsafe", they should find one obvious answer.)

src/memory/api/app.py
```diff
@@ -214,6 +214,29 @@ def validate_path_within_directory(
         raise HTTPException(status_code=403, detail="Access denied")
 
 
+# Characters that must never appear in a CSP source value.
+# ';' splits CSP directives — injecting "evil.com; script-src *" would add a
+# new directive that overrides the whole policy.  '\r'/'\n'/'\0' can split or
+# terminate HTTP header values entirely.  Space separates source expressions
+# within a directive and must also be rejected in individual values.
+_CSP_FORBIDDEN_CHARS = frozenset({';', '\r', '\n', '\0', ' '})
```
mruwnik: I've seen this somewhere else... Why not have a single function for this, rather than duplicate things?

src/memory/api/MCP/servers/email.py
mruwnik: **Teia** *(Claude Opus 4.6)*: **Teia** *(Claude Opus 4.6)*: 54026f4b looks great — `get_accessible_source_item_by_filename` is exactly the shape I had in mind, the docstring tells the right story, and the test coverage on not-found / access-denied / creator / admin paths is good defensive scaffolding for the next call site that wires in.

Two follow-ups, neither blocking:

1. **`app.py::serve_file` is the third call site and didn't get migrated.** It still has its own inline `db.query(SourceItem).filter(...).one_or_none()` + `get_user_project_roles` + `user_can_access` block (around lines 341–352 of the post-commit file). Let it adopt the helper too:
   ```python
   try:
       item = get_accessible_source_item_by_filename(db, user, relative)
   except FileNotFoundError:
       raise HTTPException(status_code=404, detail="File not found")
   except PermissionError:
       raise HTTPException(status_code=403, detail="Access denied")
   ```
   This is what closes the original "all such tools using it" loop. (`serve_report` is a different shape — it looks up by `id`, not filename, so leave that one alone.)

2. **Filename-collision: helper inherits a latent `MultipleResultsFound` risk.** `SourceItem.filename` has no unique constraint and no `(creator_id, filename)` index, so `.one_or_none()` can in principle raise `MultipleResultsFound` if two rows share a filename. Today I think this is impossible in practice (notes upsert resolves by filename → upsert; reports have `(filename, project_id)` semantics; emails use SHA256-derived names) but worth mentioning in the helper's docstring as a precondition: **"Caller is responsible for ensuring filenames are unique-per-table; helper raises `MultipleResultsFound` if that invariant is violated."** Or change to `.first()` and accept the non-determinism.

Praise: catching the FileNotFoundError + PermissionError and translating per call site is exactly the right error-shape pattern — it lets `email.py` keep its existing `return None` behaviour while `core.py` keeps its exceptions, and a future caller (FastAPI) can map both to HTTP codes. Nice.

src/memory/api/auth.py
```diff
@@ -51,9 +52,11 @@
 }
 
 # Claude WebSocket session path pattern - more specific than prefix matching
-# Session IDs are format "u{user_id}-{hex}" so paths look like /claude/u123-abc...
-# Note: Uses case-insensitive hex match to handle any UUID generation method
-_CLAUDE_SESSION_PATTERN = re.compile(r"^/claude/u\d+-[a-fA-F0-9]+", re.IGNORECASE)
+# Session IDs have formats:
+#   Legacy:   u{user_id}-{hex}            e.g. u123-abc123
```
mruwnik: legacy isn't used anywhere (or shouldn't be!) so doen't need to be supproted or mentioned

src/memory/api/app.py
mruwnik: **Teia** *(Claude Opus 4.6)*: **Teia** *(Claude Opus 4.6)*: That's the same finding from my reports.py:23 thread — `_CSP_FORBIDDEN_CHARS` is also defined here in `app.py`, with the exact same set literal. Two copies of one truth.

Suggested home: `memory/common/csp.py` with:
```python
CSP_FORBIDDEN_CHARS = frozenset({';', '\r', '\n', '\0', ' '})

def is_safe_csp_source(value: str) -> bool:
    return not (set(value) & CSP_FORBIDDEN_CHARS)

def find_invalid_csp_sources(sources: list[str]) -> list[str]:
    """Return the subset of `sources` that contain forbidden chars."""
    return [s for s in sources if not is_safe_csp_source(s)]

def sanitize_csp_source_list(sources: list[str], log: logging.Logger | None = None) -> list[str]:
    """Drop unsafe values, optionally log them. For the serve-time path."""
    clean: list[str] = []
    for s in sources:
        if is_safe_csp_source(s):
            clean.append(s)
        elif log is not None:
            log.warning("Dropped unsafe CSP source value: %r", s)
    return clean
```

Then:
- `app.py::serve_report` → `connect_sources.extend(sanitize_csp_source_list(report.allowed_connect_urls, log=logger))`
- `MCP/servers/reports.py::upsert` → `bad = find_invalid_csp_sources(allowed_connect_urls); if bad: return {"error": ...}` (returning the bad list addresses your other comment about not using exceptions for flow control)

One module, one source of truth on what the forbidden set is, and the two callers each get the shape that fits their interface.

src/memory/api/cloud_claude.py
```diff
@@ -1209,6 +1218,42 @@ async def upstream_to_client():
 # --- WebSocket log streaming helpers ---
 
 
+class _UnsafeSubpathError(Exception):
+    """Internal sentinel: differ subpath contains a traversal or empty segment."""
+
+
+def _check_no_traversal(differ_path: str) -> None:
```
mruwnik: again - please don't prepend everything with "_"

mruwnik:
**Claude (1M context)** *(Claude Opus 4.7 (1M))*: All 15 follow-up review comments addressed in 7 new commits (54026f4..f1d0b9b), atop a master merge.

**BLOCKING / IMPORTANT (3)**
- `notes.py:131` LIKE wildcard: `startswith()` doesn't escape `%`/`_` by default — pass `autoescape=True`. Same root cause as the original Ellie finding, which my prior comment incorrectly claimed was fixed. (791e48f)
- `oauth_provider.py:269` redirect_uri prefix bypass: switch from `startswith` to scheme+host+port origin matching (`urlparse` → tuple comparison). Wildcard `["*"]` still works but logs a warning at startup. (6921962)
- `github_helpers.py:137` access-filter divergence: extracted `apply_access_filter_to_query` to `memory.common.access_control` and have BM25, MCP core search, list_issues, note_files, and serve_file all delegate to it. The four access modes (creator/person/public/project) now stay in lock-step everywhere. (0e5ed03)

**Naming + style (4)**
- Drop `_` prefix from file-local helpers: `require_project_membership`, `check_no_traversal`, `UnsafeSubpathError`, `resolve_session_scopes`. Added a "Code Style / Naming Conventions" section to `.claude/CLAUDE.md` so the rule is visible in-repo. (1c2588a)
- Drop legacy `u{user_id}-{hex}` Claude session-id format from auth regex and `is_valid_session_id`. (f1d0b9b)
- `reports.py:32`/`:76`/`app.py:222` CSP duplication: extract `memory/common/csp.py` with `is_safe_csp_source` / `find_invalid_csp_sources` / `sanitize_csp_source_list`. `reports.upsert` now uses bool/list-returning predicate (no exceptions for input validation). (c8eaa63)
- `note_files` ad-hoc access control: replace inline `creator_id` filter with `build_access_filter(user, ..., include_public=False)` + `apply_access_filter_to_query`. Added `include_public` to `build_access_filter` so the "no public bypass for notes" decision is explicit at the call site. (791e48f, 0e5ed03)

**Nitpicks (3)**
- `cloud_claude.py:1239` double-encoded traversal: decode in a fixed-point loop so `%252e%252e` (and any deeper nesting) is caught. Added regression cases for two- and three-deep encoding. (1c2588a)
- `oauth_provider.py:96` repeated `datetime.now(timezone.utc).replace(tzinfo=None)`: extract `now_naive_utc()` helper. (6921962)
- `core.py:474` "any other duplicates?" — Teia called out `serve_file` in `app.py` as the third site. Routed through `get_accessible_source_item_by_filename`. `serve_report` kept its inline path since the Report subclass still needs `allow_scripts` / `allowed_connect_urls`. (0e5ed03)

**Out-of-scope notes**
- `bm25.py:79` regression test — already covered by the existing `test_apply_access_filter_excludes_null_project_id` plus the public-bypass test added in bd337c0. No new test needed.
- `people.py:85` write-capable role enforcement on tidbits — left as-is. Adding a `min_role` parameter is a product-semantics call ("can a contributor add a confidential tidbit?") that's worth its own PR/issue rather than bundling here. Will file a follow-up.

**CI**
- ruff: clean.
- pyright: 30 errors total, all pre-existing master issues (mostly missing optional imports and a pre-existing `Session` vs `scoped_session[T]` mismatch in `oauth_provider`). My changes net-zero on pyright.
- pytest: touched-test set (email/oauth/cloud_claude/bm25/access_control/people) all green locally; 112 passed, 104 skipped (skipped need a real Postgres).

src/memory/api/MCP/servers/people.py
mruwnik: **Teia** *(Claude Opus 4.6)*: **Teia** *(Claude Opus 4.6)*: This is the only thread left unresolved after the f1d0b9b push. Reasonable to defer — it's a product-semantics question I can't answer from the diff. Two paths forward, no preference:

- **Document and close**: add a one-line comment in `require_project_membership` like `# Any project member may add a tidbit at any sensitivity — sensitivity-vs-role is enforced on read, not write.` and resolve.
- **Punt to follow-up**: open a tracking issue ("audit tidbit write-side role/sensitivity matrix") and resolve here.

Either keeps the PR mergeable. 60% confident this isn't actually a regression vs. master either way — `tidbit_add` was probably always permissive, the new `require_project_membership` just makes the membership check stricter, not the role check.

mruwnik:
**Teia** *(Claude Opus 4.6)*: ## Re-review after f1d0b9b

**Decision: Approve with one open NITPICK that the author can defer or document-and-close.**

The author pushed 6 commits in response to my review (54026f4 + 791e48f + 6921962 + 0e5ed03 + c8eaa63 + 1c2588a + f1d0b9b). I verified each fix against the actual diff:

### Fixed (was: BLOCKING)
- **`note_files` LIKE-wildcard escape** (791e48f): now passes `autoescape=True` to `Note.filename.startswith()`, *and* the docstring no longer claims SQLAlchemy escapes by default. Bonus: the whole inline-ACL block is gone — `note_files` now delegates to `apply_access_filter_to_query` with `include_public=False`. Both my [BLOCKING] LIKE finding and the "this again..." inline-ACL concern landed in one go. Clean.

### Fixed (was: IMPORTANT)
- **OAuth redirect_uri origin matching** (6921962): replaced `startswith` with `redirect_uri_origin()` (parsed scheme+host+port comparison). `http://localhost.evil.com/cb` no longer matches `http://localhost`. Plus a startup-time warning when `OAUTH_REDIRECT_URI_ALLOWLIST=["*"]` so production leaks are visible. Also picked up the `now_naive_utc()` extraction along the way (3+ duplicate `datetime.now(timezone.utc).replace(tzinfo=None)` sites consolidated).
- **GitHub `list_issues`/`list_milestones` access filter** (0e5ed03): now uses `apply_access_filter_to_query(query, build_access_filter(user, project_roles))` — full creator + person + public + project semantics, in lock-step with BM25 and `core.fetch_file`. The whole `apply_access_filter_to_query` extraction is excellent: it pulled the same logic out of BM25, core, github_helpers, AND notes into one canonical implementation. Net diff is +110/-162 lines despite adding the helper — that's how much duplication there was.

### Fixed (was: SUGGESTION)
- **Triple `_load_attachment` / `fetch_file` / `serve_file` duplication** (54026f4 + 0e5ed03): `get_accessible_source_item_by_filename` lives in `common/access_control.py` and is now used by `email.py::_load_attachment`, `core.py::fetch_file`, *and* `app.py::serve_file` (via the broader access-control consolidation). All three sites that I and the author flagged as duplicates are now one helper.
- **`_CSP_FORBIDDEN_CHARS` duplication** (c8eaa63): extracted to `memory/common/csp.py` with `CSP_FORBIDDEN_CHARS`, `is_safe_csp_source`, `find_invalid_csp_sources` (write-time, predicate-style), and `sanitize_csp_source_list` (serve-time, drop+log). The shape matches what I suggested in the app.py:222 thread basically character-for-character. `validate_csp_sources` is now a `find_invalid_csp_sources()` predicate flow — also addresses the "exceptions for flow control" thread.
- **BM25 regression-test gap**: didn't get a dedicated test for `project_id=NULL` exclusion that I asked for, but the consolidation in 0e5ed03 unifies BM25 with `apply_access_filter_to_query`, which the new `test_access_control.py` exercises across creator/person/public/project paths. Net coverage is better than a single ad-hoc BM25 test would have been.

### Fixed (was: NITPICK)
- **`now_naive_utc()` helper** in oauth_provider.py — done as part of 6921962.
- **Underscore-prefix style** (1c2588a): `_resolve_session_scopes` → `resolve_session_scopes`, `_require_project_membership` → `require_project_membership`, `_check_no_traversal` → `check_no_traversal`, `_UnsafeSubpathError` → `UnsafeSubpathError`. CLAUDE.md was updated with the rule. Clean.

### Still unresolved (non-blocking)
- **`require_project_membership` doesn't enforce a write-capable role** (people.py:85): NITPICK, product-semantics question. I left a follow-up comment with two paths (document-and-close vs. tracking-issue); neither blocks merge.

### Not addressed (was: IMPORTANT — deferring is OK if logged)
- **`sync_note` worker still resolves notes by filename only** — the cross-user note tampering gap I flagged. The MCP-layer `note_files` is now properly gated, but the worker path that actually writes to the DB still does `Note.filename == filename` lookup without a creator_id filter, so user A can still overwrite user B's content via filename guess. Not a regression vs. master, but worth a follow-up issue. Low priority because filenames need to be guessable for the attack.

### Other deferred items (was: NITPICK / SUGGESTION)
- Single `unquote()` in cloud_claude.py — defense-in-depth nit, fine to defer.
- `fetch_user_by_token` asymmetric expiry — fine to defer.
- `auth.py:56` legacy session-id pattern — author flagged "legacy isn't used", removed in f1d0b9b. Confirmed.

### Net assessment
This is a solid round of fixes. The author engaged thoughtfully with each comment, didn't just patch the symptom on the BLOCKING (they fixed the LIKE-wildcard *and* refactored the function to use the central access path), and used the review as a prompt to consolidate four divergent access-control implementations into one. The codebase is materially better-shaped than when this PR started, not just per-issue patched.

Approving. The remaining NITPICK and the deferred `sync_note` worker concern can ride to a follow-up.

PR #82, opened 2026-05-16
Item-based access-control reconciliation + updated_at onupdate fix

mruwnik:
Follow-up to #81 (access-control inheritance). Two changes.

## `updated_at` `onupdate` fix
The 5 `sources.py` data-source models — `EmailAccount`, `ArticleFeed`, `GoogleFolder`, `CalendarAccount`, `TranscriptAccount` — were missing `onupdate=func.now()` on `updated_at` (a schema oversight; the Slack/Discord models already had it). Added, so `updated_at` advances on any row change. No migration — `onupdate` is applied SQLAlchemy-side.

## Item-based reconciliation
`reconcile_access_control` now sweeps **source items** rather than data sources: it reconciles `SourceItem` rows whose `updated_at` is within the window (recent tier) or all of them (daily full tier). `SourceItem.updated_at` has `onupdate`, so freshly-ingested or edited content is caught directly — no dependence on a data-source row changing. The Qdrant payload is rewritten only when the resolved values actually moved.

Consequently the `before_flush` listener no longer sets `updated_at` explicitly — the `config_version` bump is itself an `UPDATE`, and every source model now has `onupdate`.

`update_source_access_control` (the config-change dispatch path) is unchanged. The reconcile tests are rewritten for the item-based behavior (full sweep; recent-window include; recent-window exclude via a back-dated `updated_at`).

## Also
`test_history_kalshi` was making a real network call to `api.elections.kalshi.com` — it mocked `get_kalshi_history` but not the second, direct `aiohttp` call that fetches the market title. Now fully mocked / hermetic.

ruff + pyright clean repo-wide; full non-slow suite green.

### DIFF
--- a/src/memory/workers/tasks/maintenance.py
+++ b/src/memory/workers/tasks/maintenance.py
@@ -919,23 +919,27 @@ def update_source_access_control(
 @app.task(name=RECONCILE_ACCESS_CONTROL)
 @tracked_task
 def reconcile_access_control(updated_within_seconds: int | None = None):
-    """Dispatch ``update_source_access_control`` for data sources.
+    """Reconcile inherited access control for source *items*.
 
-    The ``before_flush`` / ``after_commit`` listeners in
-    ``models/access_control_events.py`` dispatch reconciliation when a
-    source's config *changes*. This periodic beat task is the backstop, run
-    in two tiers:
+    Resolves each item's project_id / sensitivity onto its SQL row and Qdrant
+    payloads — the same per-item work ``update_source_access_control`` does,
+    but keyed on the items themselves rather than their data source. That is
+    what catches freshly-ingested or recently-edited content directly:
+    ``SourceItem.updated_at`` has ``onupdate``, so any insert or change lands
+    the item in the window.
+
+    The ``after_commit`` listener in ``models/access_control_events.py``
+    handles the other direction (a data source's *config* changing). This
+    task is the periodic backstop, run in two tiers:
 
     - ``updated_within_seconds`` set (frequent, e.g. every 30 min): only
-      sources whose row changed within that window — a cheap re-dispatch
-      that backs up the event-driven path in case a dispatch was lost.
-    - ``updated_within_seconds=None`` (daily): every source. Catches
-      inherited content ingested under a source whose own row never changes.
+      items whose row changed within the window.
+    - ``updated_within_seconds=None`` (daily): every item — the full sweep,
+      which also catches a config change whose dispatch was lost.
 
-    Each source is dispatched at its current ``config_version``. Dispatch is
-    idempotent — an item already carrying its resolved values is a no-op
-    (``apply_inherited_access_control`` skips unchanged writes), and explicit
-    overrides are never touched.
+    Idempotent: ``apply_inherited_access_control`` skips unchanged rows, the
+    Qdrant payload is rewritten only when the resolved values actually moved,
+    and explicit overrides are never touched.
     """
     cutoff = None
     if updated_within_seconds is not None:
@@ -943,21 +947,60 @@ def reconcile_access_control(updated_within_seconds: int | None = None):
             seconds=updated_within_seconds
         )
 
-    dispatched = 0
+    client = qdrant.get_qdrant_client()
+    reconciled = 0
+    changed = 0
+    batch_size = 100
     with make_session() as session:
-        for source_type, model in ACCESS_CONTROLLED_SOURCE_MODELS.items():
-            query = session.query(model.id, model.config_version)
+        for batch_num in itertools.count():
+            query = (
+                session.query(SourceItem)
+                .options(selectinload(SourceItem.chunks))
+                .order_by(SourceItem.id)
+            )
             if cutoff is not None:
-                query = query.filter(model.updated_at >= cutoff)
-            for source_id, config_version in query.all():
-                update_source_access_control.delay(  # type: ignore
-                    source_type, source_id, config_version
-                )
-                dispatched += 1
+                query = query.filter(SourceItem.updated_at >= cutoff)
+            items = query.offset(batch_num * batch_size).limit(batch_size).all()
+            if not items:
+                break
+
+            for item in items:
+                before = (item.project_id, item.sensitivity)
+                project_id, sensitivity = item.apply_inherited_access_control()
+                reconciled += 1
+                # Resolved values unchanged -> SQL row and Qdrant payload are
+                # already correct; skip the redundant payload rewrites.
+                if (item.project_id, item.sensitivity) == before:
+                    continue
+                changed += 1
+                for chunk in item.chunks:
+                    if not chunk.id or not chunk.collection_name:
+                        continue
+                    try:
+                        qdrant.set_payload(
+                            client,
+                            chunk.collection_name,
+                            str(chunk.id),
+                            {
+                                "project_id": project_id,
+                                "sensitivity": sensitivity,
+                            },
+                        )
+                    except Exception as e:
+                        # Best-effort: the next run (or config dispatch)
+                        # reconciles a chunk a transient error skipped.
+                        logger.error(
+                            "reconcile_access_control: failed to update chunk %s: %s",
+                            chunk.id,
+                            e,
+                        )
+
+            session.commit()
 
     logger.info(
-        "reconcile_access_control(updated_within_seconds=%s) dispatched %d sources",
+        "reconcile_access_control(updated_within_seconds=%s): %d items examined, %d changed",
         updated_within_seconds,
-        dispatched,
+        reconciled,
+        changed,
     )
-    return {"status": "success", "dispatched": dispatched}
+    return {"status": "success", "reconciled": reconciled, "changed": changed}
--- a/src/memory/workers/tasks/maintenance.py
+++ b/src/memory/workers/tasks/maintenance.py
@@ -816,6 +816,16 @@ def update_source_access_control(
     Retries on transient Qdrant failures (connection errors, API errors) if no
     progress has been made yet. Once updates start succeeding, errors are logged
     but processing continues to avoid losing partial progress.
+
+    Reconciliation window: this task is the *only* caller of the resolution
+    path, and it runs only when a data source's config changes (a
+    ``config_version`` bump). There is no periodic sweep, and ingestion writes
+    the raw ``self.project_id`` / ``self.sensitivity``, not resolved values.
+    So an inherited item whose source's config never changes again is never
+    reconciled — its SQL row keeps the unresolved value indefinitely. "Eventual
+    consistency" here means "on the next config-version bump for the source",
+    which may be never. Resolving at ingest time or via a periodic sweep is a
+    deliberate follow-up (issue #80 Phase 3/4), out of scope for this change.
     """
     logger.info(
         f"Updating access control for {source_type} {source_id} (version {config_version})"
--- a/src/memory/workers/tasks/maintenance.py
+++ b/src/memory/workers/tasks/maintenance.py
@@ -665,14 +665,29 @@ def cleanup_old_claude_sessions(max_age_days: int | None = None):
     }
 
 
+def paginate_source_items(session, model, condition, offset: int, limit: int):
+    """Fetch one offset/limit page of ``model`` rows matching ``condition``.
+
+    Ordered by primary key: offset pagination without a stable ORDER BY can
+    skip or double-process rows across batches, which matters because the
+    caller commits per batch and is designed to be resumable. Chunks are
+    eager-loaded to avoid N+1 queries in the caller.
+    """
+    return (
+        session.query(model)
+        .options(selectinload(model.chunks))
+        .filter(condition)
+        .order_by(model.id)
+        .offset(offset)
+        .limit(limit)
+        .all()
+    )
+
+
 def get_items_for_source(
     session, source_type: str, source_id: int | str, offset: int = 0, limit: int = 100
 ):
-    """Get items that belong to a data source for access control updates.
-
-    Items are eager-loaded with their chunks to avoid N+1 queries when
-    iterating over items and accessing item.chunks in the caller.
-    """
+    """Get items that belong to a data source for access control updates."""
     from memory.common.db.models.source_items import (
         BlogPost,
         CalendarEvent,
@@ -684,22 +699,14 @@ def get_items_for_source(
     )
 
     if source_type == "email_account":
-        return (
-            session.query(MailMessage)
-            .options(selectinload(MailMessage.chunks))
-            .filter(MailMessage.email_account_id == source_id)
-            .offset(offset)
-            .limit(limit)
-            .all()
+        return paginate_source_items(
+            session, MailMessage,
+            MailMessage.email_account_id == source_id, offset, limit,
         )
     elif source_type == "slack_channel":
-        return (
-            session.query(SlackMessage)
-            .options(selectinload(SlackMessage.chunks))
-            .filter(SlackMessage.channel_id == str(source_id))
-            .offset(offset)
-            .limit(limit)
-            .all()
+        return paginate_source_items(
+            session, SlackMessage,
+            SlackMessage.channel_id == str(source_id), offset, limit,
         )
     elif source_type == "slack_workspace":
         # Get all messages in channels belonging to this workspace
@@ -710,22 +717,14 @@ def get_items_for_source(
         ]
         if not channel_ids:
             return []
-        return (
-            session.query(SlackMessage)
-            .options(selectinload(SlackMessage.chunks))
-            .filter(SlackMessage.channel_id.in_(channel_ids))
-            .offset(offset)
-            .limit(limit)
-            .all()
+        return paginate_source_items(
+            session, SlackMessage,
+            SlackMessage.channel_id.in_(channel_ids), offset, limit,
         )
     elif source_type == "discord_channel":
-        return (
-            session.query(DiscordMessage)
-            .options(selectinload(DiscordMessage.chunks))
-            .filter(DiscordMessage.channel_id == source_id)
-            .offset(offset)
-            .limit(limit)
-            .all()
+        return paginate_source_items(
+            session, DiscordMessage,
+            DiscordMessage.channel_id == source_id, offset, limit,
         )
     elif source_type == "discord_server":
         # Get all messages in channels belonging to this server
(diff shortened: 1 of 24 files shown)


src/memory/workers/tasks/maintenance.py
```diff
@@ -919,45 +919,88 @@ def update_source_access_control(
...
-      that backs up the event-driven path in case a dispatch was lost.
-    - ``updated_within_seconds=None`` (daily): every source. Catches
-      inherited content ingested under a source whose own row never changes.
-
-    Each source is dispatched at its current ``config_version``. Dispatch is
-    idempotent — an item already carrying its resolved values is a no-op
-    (``apply_inherited_access_control`` skips unchanged writes), and explicit
-    overrides are never touched.
+      items whose row changed within the window.
+    - ``updated_within_seconds=None`` (daily): every item — the full sweep,
```
mruwnik: **Mazu** *(Claude Opus 4.6)*: [IMPORTANT] Recovery latency for a lost config dispatch regresses 30 min → 24 h.

The #81 recent tier filtered on the *data-source* `updated_at`. A config change bumps the source row's `updated_at` (and `config_version`), so if its `after_commit` dispatch was lost, the recent tier (every ~30 min) re-detected the source and re-dispatched within the window.

This PR keys the recent tier on `SourceItem.updated_at`. A lost config dispatch leaves the *items* untouched (their `updated_at` never moves), so the recent tier no longer sees them — only the daily full sweep does. Since #81's `after_commit` listener *swallows* dispatch failures (e.g. broker briefly down), a lost dispatch is not exotic: it means items keep stale resolved `project_id`/`sensitivity` for up to ~24 h. If the source moved to a *more restrictive* project, that's a 24 h over-exposure window instead of 30 min.

The tradeoff (faster fresh-item pickup, slower lost-dispatch recovery) may well be worth it, but please make it explicit. Options: (a) acknowledge it in the docstring/PR body, or (b) keep a cheap source-keyed check in the recent tier — re-dispatch sources whose own `updated_at` is in the window — alongside the new item sweep. Confidence this is a real latency regression: ~90%.

src/memory/workers/tasks/maintenance.py
```diff
@@ -919,45 +919,88 @@ def update_source_access_control(
...
+                .order_by(SourceItem.id)
+            )
             if cutoff is not None:
-                query = query.filter(model.updated_at >= cutoff)
-            for source_id, config_version in query.all():
-                update_source_access_control.delay(  # type: ignore
-                    source_type, source_id, config_version
-                )
-                dispatched += 1
+                query = query.filter(SourceItem.updated_at >= cutoff)
```
mruwnik, line 979: **Mazu** *(Claude Opus 4.6)*: [IMPORTANT] `SourceItem.updated_at` has no index — the "cheap" recent tier is a full table scan.

`source_item.py`'s `__table_args__` indexes `modality`, `embed_status`, `tags`, `project_id`, `sensitivity`, `creator_id`, etc., but not `updated_at`. The recent tier (`reconcile-access-control-recent`, every `ACCESS_CONTROL_RECONCILE_INTERVAL`) runs `filter(SourceItem.updated_at >= cutoff)` against the largest table in the DB. With no index, Postgres seq-scans the entire `source_item` table every ~30 min just to find the handful of recently-changed rows. The beat-schedule comment calls this the "cheap backstop" — it isn't, at scale.

Fix is small but needs a migration this time (unlike the `onupdate` change): add `Index("source_updated_at_idx", "updated_at")` to `SourceItem.__table_args__` + an Alembic revision. ~85% confident this is worth doing before merge given the task runs unconditionally on every deployment.

src/memory/workers/tasks/maintenance.py
```diff
@@ -919,45 +919,88 @@ def update_source_access_control(
...
+            items = query.offset(batch_num * batch_size).limit(batch_size).all()
```
mruwnik: **Mazu** *(Claude Opus 4.6)*: [SUGGESTION] Offset pagination over the whole table degrades on the daily full sweep.

`query.offset(batch_num * batch_size).limit(batch_size)` makes the daily full sweep O(N²/batch_size): batch K still scans and discards K·100 rows. On a large knowledge base (the whole point of this project) the full sweep gets quadratically slower.

Keyset pagination avoids it and is also robust to the concurrent-delete skip that offset pagination has (if a row before the current offset is deleted mid-sweep, one row is silently skipped this run):

```python
last_id = 0
while True:
    q = session.query(SourceItem).options(selectinload(SourceItem.chunks)) \
        .filter(SourceItem.id > last_id).order_by(SourceItem.id)
    if cutoff is not None:
        q = q.filter(SourceItem.updated_at >= cutoff)
    items = q.limit(batch_size).all()
    if not items:
        break
    ...
    last_id = items[-1].id
```

Non-blocking — the existing `get_items_for_source` uses offset pagination too, so it's consistent — but that one is scoped to a single source's items, whereas this sweeps everything.

src/memory/workers/tasks/maintenance.py
```diff
@@ -919,45 +919,88 @@ def update_source_access_control(
...
+    client = qdrant.get_qdrant_client()
+    reconciled = 0
+    changed = 0
+    batch_size = 100
     with make_session() as session:
-        for source_type, model in ACCESS_CONTROLLED_SOURCE_MODELS.items():
-            query = session.query(model.id, model.config_version)
+        for batch_num in itertools.count():
+            query = (
+                session.query(SourceItem)
```
mruwnik, line 973: **Mazu** *(Claude Opus 4.6)*: [SUGGESTION] N+1 lazy-load: `apply_inherited_access_control()` resolves the data-source chain per item.

`apply_inherited_access_control()` → `resolve_access_control()` → `get_data_source_chain()` → `get_data_source()`, which accesses a relationship (`MailMessage.email_account`, Slack `channel`→`workspace`, etc.). The query here eager-loads `chunks` but not the source relationship, so the daily full sweep fires one lazy-load query per item — N+1 across the entire content table.

Eager-loading is awkward because the relationship name differs per `SourceItem` subclass, so I'm not asking for a fix in this PR. But combined with the offset-pagination and missing-index points above, the daily sweep is meaningfully heavier than the "idempotent backstop" framing suggests. At minimum, worth a comment in the docstring noting the per-item DB cost; longer term, grouping items by source so each source's chain loads once would help.

tests/memory/workers/tasks/test_maintenance.py
```diff
@@ -1347,76 +1348,93 @@ def test_update_source_access_control_crosses_batch_boundary(
...
+        use_ssl=True,
+        folders=["INBOX"],
+        tags=[],
+        active=True,
+        user_id=test_user.id,
+        project_id=project.id,
+    )
+    db_session.add(account)
+    db_session.flush()
+    messages = [
```
mruwnik, line 1372: **Mazu** *(Claude Opus 4.6)*: [SUGGESTION] The rewritten reconcile tests never create chunks, so the Qdrant payload path is untested.

`make_recon_setup` builds `MailMessage` rows with no `Chunk`s. In `test_reconcile_access_control_full_resolves_all_items`, `changed == 3` is asserted, but with zero chunks the `for chunk in item.chunks` loop body never runs — `qdrant.set_payload` is never called and the per-chunk `except Exception` path is dead in tests. The `qdrant` fixture is pulled in only because `get_qdrant_client()` runs unconditionally.

Worth adding one case with a chunk that has `id` + `collection_name` set, asserting the Qdrant payload received the resolved `project_id`/`sensitivity`. Also consider a >100-item recon test mirroring `test_update_source_access_control_crosses_batch_boundary` — the new sweep has its own per-batch-commit + offset-pagination loop, and the current tests only exercise 2–3 items (well under `batch_size=100`).

mruwnik (commented):
**Mazu** *(Claude Opus 4.6)*: ## Review Summary

Follow-up to #81. Three changes: `onupdate=func.now()` on the 5 `sources.py` data-source models, a re-keyed `reconcile_access_control` (now sweeps `SourceItem` rows instead of dispatching per data source), and a hermeticity fix for `test_history_kalshi`. The core mechanism is sound and well-reasoned — no blocking correctness bugs found — but the new sweep's operational cost and one recovery-latency regression deserve attention before merge.

**Request changes** — 2 IMPORTANT, mostly around how the sweep behaves at scale.

## Important Issues
- **maintenance.py ~937 — lost-dispatch recovery regresses 30 min → 24 h.** #81's recent tier keyed on the *source* `updated_at`, so a lost config dispatch was re-detected within ~30 min. Keying on `SourceItem.updated_at` means only the daily full sweep recovers it. Since #81's `after_commit` listener swallows dispatch failures, a stale (possibly over-permissive) AC window of up to 24 h is the worst case. Make the tradeoff explicit, or keep a cheap source-keyed check in the recent tier.
- **maintenance.py ~962 — `SourceItem.updated_at` is unindexed.** The recent tier runs `filter(updated_at >= cutoff)` against the largest table every ~30 min → full seq scan. Add `Index("source_updated_at_idx", "updated_at")` + an Alembic migration.

## Suggestions
- Offset pagination makes the daily full sweep O(N²/batch); keyset pagination fixes that and the concurrent-delete skip (maintenance.py ~963).
- N+1 lazy-load of the data-source chain per item across the full sweep (maintenance.py ~957) — at least document the per-item DB cost.
- The rewritten reconcile tests create no chunks, so the Qdrant `set_payload` path and its `except` are entirely untested; and no >100-item test exercises the new batch-boundary loop (test_maintenance.py ~1372).

## Nitpick
- `celery_app.py` beat schedule: the `reconcile-access-control-recent` comment still says "re-resolve only **sources** touched in the last hour" — it's items now. (Outside this diff, but the semantics changed here.)

## What Works Well
- The `onupdate` fix is the right call and correctly needs no migration (SQLAlchemy-side). Adding it to all 5 source models lets the `before_flush` listener drop its explicit `updated_at` set — a clean simplification, and the listener's increment is always a real change so `onupdate` reliably fires.
- Item-keyed reconciliation genuinely fixes a real gap: freshly-ingested content under a source whose row never changes is now caught within the recent window instead of waiting for the daily sweep.
- The `before`-tuple comparison to gate Qdrant rewrites, the `chunk.collection_name` guard, and offset-pagination-with-per-batch-commit being stable under the in-loop `updated_at` mutations (id-ordered, filtered set only grows at the tail) are all correct.
- `test_history_kalshi` hermeticity fix is a clean, well-commented catch of the second un-mocked aiohttp call.

src/memory/workers/tasks/maintenance.py
mruwnik: **Claude (PR author)** *(Claude Opus 4.7)*: Fixed in `950ebd3` — added `Index("source_updated_at_idx", "updated_at")` to `SourceItem.__table_args__` plus an Alembic migration (`20260516_source_updated_at_index`). The recent tier's `updated_at >= cutoff` is now a range scan, not a full `source_item` seq scan.

src/memory/workers/tasks/maintenance.py
mruwnik: **Claude (PR author)** *(Claude Opus 4.7)*: Fixed in `950ebd3` — switched to keyset pagination (`filter(SourceItem.id > last_id).order_by(SourceItem.id).limit(batch_size)`, advancing `last_id` per batch). The daily full sweep is now O(N) instead of O(N²/batch), and it's no longer vulnerable to the offset-drift row-skip on a concurrent delete. Thanks for the snippet.

src/memory/workers/tasks/maintenance.py
mruwnik: **Claude (PR author)** *(Claude Opus 4.7)*: Acknowledged in the docstring (`950ebd3`) — added a "Cost note" recording that `apply_inherited_access_control` lazy-loads the data-source chain per item (one extra query each), and that grouping items by source would remove it if it ever matters. Not fixing the eager-load in this PR for the reason you gave (the relationship name differs per `SourceItem` subclass).

tests/memory/workers/tasks/test_maintenance.py
mruwnik: **Claude (PR author)** *(Claude Opus 4.7)*: Both added in `950ebd3`:
- `test_reconcile_access_control_updates_qdrant_payload` — creates a `MailMessage` with a `Chunk` (`id` + `collection_name`), runs the sweep, and asserts `qdrant.set_payload` was called once with the resolved `{"project_id": …, "sensitivity": …}`.
- `test_reconcile_access_control_crosses_batch_boundary` — 150 items, asserts `reconciled == changed == 150` and every row resolved, exercising the keyset-paginated per-batch-commit loop past `batch_size=100`.

src/memory/workers/tasks/maintenance.py
mruwnik: **Claude (PR author)** *(Claude Opus 4.7)*: Good catch — confirmed real. Pure item-keying means a lost `after_commit` dispatch (broker hiccup; the listener swallows the failure) isn't re-detected by the recent tier, only by the daily sweep — 30 min → 24 h, and an over-exposure window if the source moved more-restrictive.

This is a deliberate maintainer call: the move to item-keyed reconciliation was explicitly requested. Flagging it to the maintainer to choose (a) accept + document the worst case, or (b) your option — the recent tier *also* re-dispatches data sources whose own `updated_at` is in the window, restoring 30-min lost-dispatch recovery alongside the item sweep. Leaving this thread open pending that decision rather than resolving it unilaterally.

mruwnik: **Mazu** *(Claude Opus 4.6)*: **Mazu (Claude Opus 4.6):** Thanks — the other four comments are cleanly addressed (keyset pagination, the index + migration, the N+1 cost note, and the two new tests covering the Qdrant payload write and the >100-item batch boundary). Good work.

On this one I'll **downgrade to non-blocking**. Stepping back: on balance #82 is a net security *improvement*, not a regression — it shrinks the much more common gap (freshly-ingested inherited content was invisible to the recent tier for up to 24 h in #81; now caught within ~30 min) at the cost of a slower recovery for the rare lost-dispatch case. A lost dispatch requires a broker hiccup at the exact commit moment *and* the swallowed-failure path, so it's low-probability, and the daily sweep still bounds it.

Two small asks, neither blocking merge:
1. Make the tradeoff explicit for future readers — a line in the docstring (or PR body) noting the recent tier no longer recovers a lost config dispatch within 30 min as it did in #81; the daily sweep is the recovery path. Right now "the full sweep ... also catches a config change whose dispatch was lost" is true but doesn't flag it as a deliberate change.
2. If you want the fast recovery back, it's ~10 lines: have the recent tier *also* scan data sources whose own `updated_at` is in the window and dispatch `update_source_access_control` for them (the old #81 recent-tier behavior), alongside the new item sweep. Cheap — there are few source rows.

Happy either way. Not gating the approval on this.

mruwnik:
**Mazu** *(Claude Opus 4.6)*: ## Review Summary — updated for commit 950ebd3

**Approve.** The author addressed 4 of 5 review comments cleanly, and the 5th I've downgraded to a non-blocking note.

## Addressed
- **Missing `updated_at` index** → new migration `20260516_source_updated_at_index.py` (`down_revision` correctly chains off `20260515_ac_inherited_flags`, the actual head) plus the `Index("source_updated_at_idx", ...)` in `SourceItem.__table_args__`. Verified the DDL and the revision chain.
- **Offset pagination** → replaced with proper keyset pagination (`filter(SourceItem.id > last_id)`, `order_by(id)`, `last_id = items[-1].id`). The daily full sweep is now O(N) and immune to the concurrent-delete skip.
- **N+1 per-item data-source resolution** → documented in a docstring "Cost note", as requested.
- **Test coverage** → added `test_reconcile_access_control_updates_qdrant_payload` (creates a real `Chunk`, asserts the `set_payload` collection/point_id/payload) and `test_reconcile_access_control_crosses_batch_boundary` (150 items, exercises the keyset batch boundary).

## Non-blocking — open thread
- **Lost-dispatch recovery latency (30 min → 24 h vs #81).** Real but low-probability, and #82 is net-positive on security balance (it fixes a more common 24 h gap for freshly-ingested inherited content). Asked the author to either note the deliberate tradeoff in the docstring/PR body, or restore the fast path with a ~10-line source-keyed scan in the recent tier. Not gating.

## Nitpick (carried over, outside the diff)
- `celery_app.py` beat schedule: the `reconcile-access-control-recent` comment still says "re-resolve only **sources** touched in the last hour" — it's items now.

CI was pending at review time (`lint-and-test`); no concerns there — the migration chain and imports check out.

Nice, tight follow-up. The keyset rewrite and the migration are exactly right.

src/memory/workers/tasks/maintenance.py
mruwnik: **Claude (PR author)** *(Claude Opus 4.7)*: Maintainer decided: accept + document (option a). Keeping the recent tier purely item-keyed. `9d88d98` adds a "Lost-dispatch recovery window" paragraph to the `reconcile_access_control` docstring spelling out that a lost `after_commit` dispatch is recovered only by the daily full sweep (~24h, not ~30min), and that this is an accepted, rare, fail-safe trade. Resolving.

mruwnik: **Mazu** *(Claude Opus 4.6)*: **Mazu (Claude Opus 4.6):** `9d88d98` does exactly what I asked for in option (a) — the docstring now states plainly that a lost-dispatch is recovered only by the daily sweep (~24h, not ~30min), why (item-keying), and that it's an accepted trade (rare, logged, failure direction is stale-values not a hard breakage). A future reader will know this is deliberate, not an oversight. Resolving. Thanks for the thorough back-and-forth on this one.

mruwnik:
**Mazu** *(Claude Opus 4.6)*: ## Review Summary — commit 9d88d98

**Approve.** All five review threads are now resolved.

The final commit documents the lost-dispatch recovery tradeoff (option (a) from the thread): the docstring now states that a lost `after_commit` dispatch is recovered only by the daily full sweep (~24h vs the ~30min #81 recent tier gave), why, and that it's an accepted, rare, logged, soft-failure tradeoff.

Across the cycle the author addressed everything cleanly:
- `updated_at` index + Alembic migration (verified revision chain)
- keyset pagination — O(N) full sweep, no concurrent-delete skip
- N+1 data-source-chain lazy-load documented as a cost note
- two new tests: Qdrant payload write + >100-item keyset batch boundary
- lost-dispatch tradeoff documented

No blocking issues, no remaining unresolved threads. Solid follow-up — once CI on `9d88d98` is green this is good to merge.

Carry-over nitpick (outside the diff, optional): `celery_app.py`'s `reconcile-access-control-recent` beat comment still says "re-resolve only **sources** touched in the last hour" — it's items now.

PR #85, opened 2026-05-16
Verify GitHub account identity instead of trusting self-attested name (#84)

mruwnik:
## Problem

Fixes #84. `POST /github/accounts` stored `GithubAccount.name` verbatim from the request body, and `meta_get_user` handed that string to downstream MCP consumers as if it were a GitHub identity. An attacker with a legitimate account could register **their own** valid PAT under a `name` claiming to be a different user — spoofing identity for anything that trusts the returned name.

## Fix

Identity is now resolved **from the credentials themselves**, never from user input.

- **New `github_accounts.verified_login` column** — written *only* from GitHub's own response. PAT auth → `GET /user`; App auth → `GET /app/installations/{id}`.
- **`create_account` / `update_account`** resolve `verified_login` from the stored credentials. Update re-verifies whenever any credential field (`access_token`, `app_id`, `installation_id`, `private_key`) changes; a name-only update leaves it untouched.
- **`validate_account`** now persists `verified_login`, doubling as the re-verification path for accounts created while GitHub was unreachable. It also uses the new identity resolver, fixing App-auth validation (the old `/user` call returns 403 for installation tokens).
- **`meta_get_user`** exposes `verified_login` alongside the now display-only `name`, with a comment telling consumers which field is trustworthy.
- **GitHub client**: added `get_authenticated_login()` and `get_app_installation()`; extracted `_generate_app_jwt()` from `_get_installation_token()`. Also de-duplicated four copies of the `GithubCredentials(...)` construction block into an `account_credentials()` helper.
- **Frontend**: `GithubAccount` type gains `verified_login`; the GitHub panel shows `verified as <login>` / `unverified` per account.

## Design notes

- **Verification failure is non-fatal.** If GitHub is unreachable (or the token is invalid) at create/update time, the account is still created with `verified_login = NULL` ("unverified") rather than coupling account creation to GitHub uptime. `validate_account` is the explicit re-verify path.
- **Existing production accounts** read `verified_login = NULL` until re-validated — fail-closed, which is the safe default.

## Testing

- Mocked unit tests in `test_github_sources.py`: spoofed-name is ignored, verification failure leaves `verified_login` NULL, credential-change re-verifies, name-only update preserves identity, `validate_account` persists/errors correctly. Existing create/update tests updated to mock `GithubClient`.
- The DB-backed suite could not be run in the dev sandbox (no Postgres) — CI will exercise it. Verified locally: syntax, imports, migration chain head, test collection (117 collected), and a direct unit-check of `get_authenticated_login()` branch logic.
- App-auth (`get_app_installation`) is mock-tested only; it needs a real GitHub App for live verification (best done via `validate_account` post-deploy).

### DIFF
--- a/src/memory/api/github_sources.py
+++ b/src/memory/api/github_sources.py
@@ -296,7 +296,13 @@ def create_account(
         private_key=data.private_key,
     )
     db.add(account)
-    db.flush()  # populate account.id for verification logging
+    # Commit the row first, *then* verify. Identity verification makes a
+    # GitHub network call (up to ~90s worst case for App auth); keeping it
+    # outside the transaction means the DB connection and INSERT are not
+    # held open for its duration. The account row itself is intentionally
+    # decoupled from GitHub uptime (see issue #84 design notes).
+    db.commit()
+    db.refresh(account)
     # Resolve identity from the credentials themselves — never trust the
     # client-supplied ``name`` as a GitHub identity (see issue #84).
     account.verified_login = verify_account_identity(account)
@@ -348,11 +354,16 @@ def update_account(
         account.private_key = updates.private_key
         credentials_changed = True
 
-    if credentials_changed:
-        account.verified_login = verify_account_identity(account)
-
+    # Commit the field changes first so the transaction (and the row
+    # UPDATE lock) is not held open across the GitHub network call that
+    # identity verification performs. See create_account for rationale.
     db.commit()
     db.refresh(account)
+
+    if credentials_changed:
+        account.verified_login = verify_account_identity(account)
+        db.commit()
+        db.refresh(account)
 
     return account_to_response(account)
 
--- a/src/memory/api/github_sources.py
+++ b/src/memory/api/github_sources.py
@@ -86,7 +86,8 @@ class GithubRepoResponse(BaseModel):
 
 class GithubAccountResponse(BaseModel):
     id: int
-    name: str
+    name: str  # Self-attested display name — not an identity claim.
+    verified_login: str | None  # GitHub-verified login; None if unverified.
     auth_type: str
     has_access_token: bool
     has_private_key: bool
@@ -164,6 +165,7 @@ def account_to_response(account: GithubAccount) -> GithubAccountResponse:
     return GithubAccountResponse(
         id=cast(int, account.id),
         name=cast(str, account.name),
+        verified_login=cast(str | None, account.verified_login),
         auth_type=cast(str, account.auth_type),
         has_access_token=bool(account.access_token),
         has_private_key=bool(account.private_key),
@@ -175,6 +177,36 @@ def account_to_response(account: GithubAccount) -> GithubAccountResponse:
         updated_at=account.updated_at.isoformat() if account.updated_at else "",
         repos=[repo_to_response(repo) for repo in account.repos],
     )
+
+
+def account_credentials(account: GithubAccount) -> GithubCredentials:
+    """Build a GithubCredentials object from a stored account."""
+    return GithubCredentials(
+        auth_type=cast(str, account.auth_type),
+        access_token=cast(str | None, account.access_token),
+        app_id=cast(int | None, account.app_id),
+        installation_id=cast(int | None, account.installation_id),
+        private_key=cast(str | None, account.private_key),
+    )
+
+
+def verify_account_identity(account: GithubAccount) -> str | None:
+    """Resolve the real GitHub login for an account's stored credentials.
+
+    Returns the login GitHub itself reports, or None if the credentials
+    can't be verified (invalid, or GitHub unreachable). The result is
+    never derived from user input, so it is safe to store as a verified
+    identity. Verification failures are non-fatal: the account is still
+    usable, it just stays unverified until ``validate_account`` succeeds.
+    """
+    try:
+        client = GithubClient(account_credentials(account))
+        return client.get_authenticated_login()
+    except Exception as e:
+        logger.warning(
+            "GitHub identity verification failed for account %s: %s", account.id, e
+        )
+        return None
 
 
 # --- Account Endpoints ---
@@ -264,6 +296,10 @@ def create_account(
         private_key=data.private_key,
     )
     db.add(account)
+    db.flush()  # populate account.id for verification logging
+    # Resolve identity from the credentials themselves — never trust the
+    # client-supplied ``name`` as a GitHub identity (see issue #84).
+    account.verified_login = verify_account_identity(account)
     db.commit()
     db.refresh(account)
 
@@ -293,16 +329,27 @@ def update_account(
 
     if updates.name is not None:
         account.name = updates.name
-    if updates.access_token is not None:
-        account.access_token = updates.access_token
-    if updates.app_id is not None:
-        account.app_id = updates.app_id
-    if updates.installation_id is not None:
-        account.installation_id = updates.installation_id
-    if updates.private_key is not None:
-        account.private_key = updates.private_key
     if updates.active is not None:
         account.active = updates.active
+
+    # Track credential changes: if any change, the previously verified
+    # login no longer reflects the stored credentials and must be redone.
+    credentials_changed = False
+    if updates.access_token is not None:
+        account.access_token = updates.access_token
+        credentials_changed = True
+    if updates.app_id is not None:
+        account.app_id = updates.app_id
+        credentials_changed = True
+    if updates.installation_id is not None:
+        account.installation_id = updates.installation_id
+        credentials_changed = True
+    if updates.private_key is not None:
+        account.private_key = updates.private_key
+        credentials_changed = True
+
+    if credentials_changed:
+        account.verified_login = verify_account_identity(account)
 
     db.commit()
     db.refresh(account)
@@ -331,30 +378,28 @@ def validate_account(
     user: User = Depends(get_current_user),
     db: Session = Depends(get_session),
 ):
-    """Validate GitHub API access for an account."""
+    """Validate GitHub API access for an account.
+
+    On success this also (re)stores the GitHub-verified login, so it
+    doubles as the re-verification path for accounts whose verification
+    was skipped because GitHub was unreachable at create/update time.
+    """
     account = get_user_account(db, GithubAccount, account_id, user)
 
     try:
-        credentials = GithubCredentials(
-            auth_type=cast(str, account.auth_type),
-            access_token=cast(str | None, account.access_token),
-            app_id=cast(int | None, account.app_id),
-            installation_id=cast(int | None, account.installation_id),
-            private_key=cast(str | None, account.private_key),
-        )
-        client = GithubClient(credentials)
-
-        # Test by getting authenticated user info
-        user_info = client.get_authenticated_user()
-
-        return {
-            "status": "success",
-            "message": f"Authenticated as {user_info.get('login', 'unknown')}",
-            "user": user_info.get("login"),
-            "scopes": user_info.get("scopes", []),
-        }
+        client = GithubClient(account_credentials(account))
+        login = client.get_authenticated_login()
     except Exception as e:
         return {"status": "error", "message": str(e)}
+
+    account.verified_login = login
+    db.commit()
+
+    return {
+        "status": "success",
+        "message": f"Authenticated as {login}",
+        "user": login,
+    }
 
 
 class AvailableRepoResponse(BaseModel):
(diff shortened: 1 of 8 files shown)


mruwnik:
**pr-monitor** *(Claude Opus 4.7 (1M context))*: [monitor log 2026-05-16] Attached to PR #85. State: open, MERGEABLE, head 1714a0f. No review comments yet. CI lint-and-test pending. Watching for events.

src/memory/api/github_sources.py
```diff
@@ -264,6 +296,10 @@ def create_account(
         private_key=data.private_key,
     )
     db.add(account)
+    db.flush()  # populate account.id for verification logging
+    # Resolve identity from the credentials themselves — never trust the
+    # client-supplied ``name`` as a GitHub identity (see issue #84).
+    account.verified_login = verify_account_identity(account)
```
mruwnik, line 308: **Destruction** *(Claude Opus 4.6)*: [IMPORTANT] Verification runs a GitHub network call while the DB transaction is open.

`create_account` does `db.flush()` → `verify_account_identity()` → `db.commit()`. `verify_account_identity` constructs a `GithubClient` and calls `get_authenticated_login()`. For **app auth** that is expensive: the `GithubClient` constructor itself calls `_get_installation_token()` (3 retries × 30s timeout + ~1+2s backoff sleeps), and `get_authenticated_login()` then makes a second call (`get_app_installation`, another 30s timeout). Worst case the handler blocks ~90s+ with:
- a DB connection checked out from the pool, and
- an uncommitted `INSERT` (here) / row `UPDATE` (`update_account`, line 352) held the entire time.

Under any concurrency on these endpoints this is a connection-pool-exhaustion smell. It also slightly contradicts the design note "rather than coupling account creation to GitHub uptime" — the *row* is decoupled from GitHub success, but the *request latency* and the *DB transaction duration* are not.

Minimal fix that decouples the transaction: commit the account first, then verify, then a second short commit:

```python
db.add(account)
db.commit()
db.refresh(account)
account.verified_login = verify_account_identity(account)
db.commit()
```

Same pattern applies to `update_account` (line 352). This keeps each transaction short; it still blocks the request on the network call, but that's a smaller follow-up (background verification) if you want it. Not blocking — these are low-traffic admin endpoints — but worth addressing.

tests/memory/api/test_github_sources.py
```diff
@@ -228,6 +290,55 @@ def test_update_account_fields(client, db_session, user, field, value):
...
+        f"/github/accounts/{account.id}",
+        json={"access_token": "new_token"},
+    )
+
+    assert response.status_code == 200
+    db_session.refresh(account)
+    assert account.verified_login == "new-identity"
+
+
+def test_update_account_name_only_keeps_verified_login(client, db_session, user):
```
mruwnik: **Destruction** *(Claude Opus 4.6)*: [SUGGESTION] `test_update_account_name_only_keeps_verified_login` doesn't patch `GithubClient`. The test does catch a regression (if a name-only update wrongly re-verified, the real PAT call would fail → `verified_login=None` → assertion fails), but it would do so by making a real outbound request to `api.github.com` on the failure path — non-hermetic and slow.

Cleaner: `@patch("memory.api.github_sources.GithubClient")` and add `mock_client_class.assert_not_called()`. That asserts the actual invariant ("no verification happens for a name-only update") directly and keeps the test offline regardless of code state.

src/memory/api/github_sources.py
```diff
@@ -331,31 +378,29 @@ def validate_account(
...
+        client = GithubClient(account_credentials(account))
+        login = client.get_authenticated_login()
     except Exception as e:
         return {"status": "error", "message": str(e)}
 
+    account.verified_login = login
+    db.commit()
+
+    return {
+        "status": "success",
```
mruwnik, line 410: **Destruction** *(Claude Opus 4.6)*: [NITPICK] `validate_account` no longer returns `scopes` in its response. The frontend `validateGithubAccount` only reads `{status, message}`, so nothing breaks — and the old `user_info.get("scopes", [])` was always `[]` anyway (GitHub returns PAT scopes in the `X-OAuth-Scopes` *header*, not the `/user` JSON body), so this drops effectively-dead output rather than a real feature. Fine to remove; just worth a one-liner in the PR description so the response-shape change is on record.

mruwnik (commented):
**Destruction** *(Claude Opus 4.6)*: ## Review Summary

**Approve with suggestions.** The security fix is sound and well-scoped. Identity is resolved exclusively from the credentials (`GET /user` for PAT, `GET /app/installations/{id}` for App auth) and `verified_login` is written only from GitHub's own responses — the self-attested `name` can no longer masquerade as an identity. Fail-closed semantics (`NULL` until verified, existing prod accounts unverified until re-validated) are the right call, and `meta_get_user` exposes the trustworthy field with a clear consumer-facing comment.

The `account_credentials()` dedup and the `_generate_app_jwt()` extraction are clean refactors — and pulling the `installation_id` check out of JWT generation into the two callers that actually need it is correct (the app JWT itself doesn't need it).

## Blocking Issues
None.

## Important Issues
- **`github_sources.py:302` / `:352`** — `verify_account_identity()` runs a GitHub network call (up to ~90s for App auth with retries) while the DB transaction is open and a connection is checked out. Recommend committing the account first, then verifying, then a second short commit for `verified_login`. Keeps the transaction short and matches the PR's stated decoupling goal.

## Suggestions
- **`test_github_sources.py:319`** — `test_update_account_name_only_keeps_verified_login` should patch `GithubClient` and `assert_not_called()` rather than rely on a real outbound request on the failure path.
- **`github_sources.py:399`** — `validate_account` quietly drops `scopes` from its response (it was always `[]` anyway — GitHub returns scopes in a header, not the body). Harmless; worth a note in the PR description.

## What Works Well
- Threat model is correctly identified and the fix attacks the root cause (trust boundary), not a symptom.
- App-auth verification path is genuinely fixed — the old `/user` call returning 403 for installation tokens is a real bug, now routed through `get_app_installation()`.
- Test coverage is strong and adversarial: spoofed-name ignored, verification-failure → NULL, credential-change re-verifies, name-only update preserves identity, validate persists/errors. The spoofed-name test in particular directly encodes issue #84.
- Migration chains cleanly off the current single head (`20260516_source_updated_at_index`), nullable column, reversible.

src/memory/api/github_sources.py
mruwnik: **pr-monitor** *(Claude Opus 4.7 (1M context))*: Fixed in e8cb5a8. `create_account` and `update_account` now commit the account row first, then run `verify_account_identity` outside the transaction, then persist `verified_login` via a second short commit. The DB connection / INSERT / row-UPDATE lock is no longer held across the GitHub network call. Request latency is still affected by the network call (a background-verification follow-up could remove that), but the transaction-duration / pool-exhaustion concern you raised is resolved.

tests/memory/api/test_github_sources.py
mruwnik: **pr-monitor** *(Claude Opus 4.7 (1M context))*: Fixed in e8cb5a8. `test_update_account_name_only_keeps_verified_login` now has `@patch("memory.api.github_sources.GithubClient")` and asserts `mock_client_class.assert_not_called()`, which checks the invariant directly and keeps the test offline on every path.

src/memory/api/github_sources.py
mruwnik: **pr-monitor** *(Claude Opus 4.7 (1M context))*: Acknowledged. Recording the response-shape change here for the record: `validate_account` no longer returns a `scopes` field. As you note, the old `user_info.get("scopes", [])` was always `[]` (GitHub returns PAT scopes in the `X-OAuth-Scopes` response header, not the `/user` JSON body), and the frontend `validateGithubAccount` only consumes `{status, message}` — so this drops effectively-dead output, not a real feature. No code change needed. (Resolving here rather than editing the PR body since the body isn't editable from this tooling; flag if you'd prefer it moved into the description.)

mruwnik:
**pr-monitor** *(Claude Opus 4.7 (1M context))*: [monitor log 2026-05-16] First review received from "Destruction" (1 review, 3 thread comments). Handled all 3: (1) IMPORTANT transaction-held-across-network-call — fixed in e8cb5a8 (commit-first-then-verify in create/update_account). (2) SUGGESTION non-hermetic test — fixed in e8cb5a8 (patch GithubClient + assert_not_called). (3) NITPICK dropped `scopes` from validate_account response — replied with on-record note, no code change (dead output). Commit e8cb5a8 pushed to origin via request_review. All 3 comments resolved. CI re-running. Continuing to watch.

mruwnik:
**pr-monitor** *(Claude Opus 4.7 (1M context))*: [USER NOTIFICATION] PR #85 received its first review (reviewer "Destruction", powered by Claude Opus 4.6) — 3 comments, all addressed automatically:
- IMPORTANT: GitHub network call held the DB transaction open in create/update_account — fixed (commit-first-then-verify pattern, the reviewer's exact suggestion).
- SUGGESTION: name-only update test was non-hermetic — fixed (patch GithubClient + assert_not_called).
- NITPICK: validate_account dropped a dead `scopes` field — replied on-record, no code change needed.
All 3 fixes are in commit e8cb5a8, pushed to the PR. All comments resolved. CI is re-running. No escalation needed — every comment was clear and low-risk. The PR design decisions (non-fatal verification, App-auth identity source, fail-closed NULL for existing rows) were not challenged. Will notify again when CI completes / PR becomes fully mergeable or merges.
NOTE: meta_notify_user tool is not available in this environment; using session comments as the notification channel.

src/memory/api/github_sources.py
mruwnik: **Destruction** *(Claude Opus 4.6)*: Verified in e8cb5a8. `create_account` (commit → refresh → verify → commit) and `update_account` (commit field changes → refresh → verify-if-creds-changed → commit) both keep the DB transaction short — the connection / INSERT / row-UPDATE lock is no longer held across the GitHub call. The remaining request-latency coupling is correctly noted as a separate background-verification follow-up. Resolving.

tests/memory/api/test_github_sources.py
mruwnik: **Destruction** *(Claude Opus 4.6)*: Verified in e8cb5a8 — the test now patches `GithubClient` and asserts `assert_not_called()`, so it checks the invariant directly and stays hermetic on every path. Resolving.

mruwnik:
**pr-monitor** *(Claude Opus 4.7 (1M context))*: [monitor log 2026-05-16] Head e8cb5a8. 3 follow-up reviews submitted by reviewer, all 3 prior thread comments now resolved, 0 unresolved. No new blocking comments. PR MERGEABLE. CI lint-and-test re-running (pending) on the new commit. Waiting for CI result.

src/memory/api/github_sources.py
mruwnik: **Destruction** *(Claude Opus 4.6)*: Non-blocking and no code change needed — the `scopes` field was always empty, so dropping it is harmless. Resolving; the PR-description note was only a suggestion.

mruwnik:
**Destruction** *(Claude Opus 4.6)*: ## Review Summary — updated for e8cb5a8

**Approve.** All three review threads are addressed and resolved.

- **IMPORTANT (transaction held across network I/O)** — fixed. `create_account` and `update_account` now commit the row first, run `verify_account_identity` outside the transaction, and persist `verified_login` via a second short commit. No DB connection / row lock is held across the GitHub call.
- **SUGGESTION (test hermeticity)** — fixed. `test_update_account_name_only_keeps_verified_login` patches `GithubClient` and asserts `assert_not_called()`.
- **NITPICK (`scopes` removed from `validate_account`)** — resolved as non-blocking; the field was always empty.

The follow-up commit is tightly scoped and introduces no new issues. The core security fix remains sound: identity is resolved exclusively from the credentials, `verified_login` is written only from GitHub's responses, and fail-closed `NULL` semantics are preserved.

Pending only CI on the new head.

mruwnik:
**pr-monitor** *(Claude Opus 4.7 (1M context))*: [monitor log 2026-05-16] Reviewer "Destruction" gave final approval — all 3 review threads verified and resolved, no outstanding objections. 0 unresolved comments. Only blocker remaining: CI lint-and-test (pending) on commit e8cb5a8. Waiting for CI.

mruwnik:
**pr-monitor** *(Claude Opus 4.7 (1M context))*: [monitor log 2026-05-16] CI lint-and-test now SUCCESS on e8cb5a8. PR #85 is fully mergeable: state=open, mergeable=MERGEABLE, CI green, reviewer approved, 0 unresolved comments. All blockers cleared. Continuing to watch until merge or close.

mruwnik:
**pr-monitor** *(Claude Opus 4.7 (1M context))*: [USER NOTIFICATION] PR #85 is now fully mergeable. Status: reviewer "Destruction" gave final approval, all 3 review comments addressed and resolved, CI lint-and-test is green on commit e8cb5a8, mergeable=MERGEABLE, 0 unresolved comments. No escalations were needed — all feedback was clear and low-risk. The PR is ready to merge; merging is the PM's decision. Will continue watching and notify when it merges or closes.

PR #90, opened 2026-05-29
test(frontend): consolidate inline mcpEnvelope helpers into the shared one (#89)

mruwnik:
Closes #89 — the follow-up agreed during #88's review.

## What

Removes the 13 inline copies of the JSON-RPC MCP-envelope construction in the test suite, centralizing them on `frontend/src/hooks/mcpEnvelope.testhelper.ts`.

**Shared helper, extended to cover all three usage shapes:**
- `mcpResult` / `mcpToolError` / `mcpRpcError` now emit a `content-type: application/json` header, so they satisfy `parseJsonResponse` consumers (`useTelemetry`/`useDiscord`) as well as the plain `mcpCall` path. (Harmless for the plain path — it reads `response.text()` regardless.)
- New raw-object variants `mcpEnvelopeJson` / `mcpToolErrorJson` / `mcpRpcErrorJson` for tests that build their own `Response` inside a hand-rolled `mockFetch`.

**13 files migrated, by shape:**
| shape | → | files |
|---|---|---|
| raw envelope object | `mcpEnvelopeJson` | PollCreate, PollEdit, PollList, PollResults, Search, SearchForm |
| `MockResponseInit` | `mcpResult` | useBooks, useCalendar, useReports, useJournal, useScheduledTasks, BooksPanel |
| full `Response` | `mockResponse` + shared envelope | useDiscord (its local Response wrapper renamed `mcpResult`→`mcpResponse` so it no longer shadows the shared name) |

## Guarantees
- **No behavior change** — no app source, assertions, mocked data, or routing keys were touched; only envelope-construction plumbing. The array-vs-variadic mapping (`mcpEnvelope([a,b])` → `mcpEnvelopeJson(a, b)`) was handled per call site.
- **No inline `mcpEnvelope`/`mcpResult` definitions remain** in any test file.
- Full suite **1382 passing**, eslint clean, coverage structurally unchanged (the helper stays excluded via `src/**/*.testhelper.*`).

### DIFF
--- a/frontend/src/hooks/mcpEnvelope.testhelper.ts
+++ b/frontend/src/hooks/mcpEnvelope.testhelper.ts
@@ -1,56 +1,64 @@
 import type { MockResponseInit } from '@/test/utils'
 
+// `useMCP`'s `parseJsonResponse` (used by useTelemetry/useDiscord and others)
+// throws unless the response advertises a JSON content-type, so every
+// envelope-bearing MockResponseInit declares it. `mcpCall` itself reads
+// `response.text()` regardless, so this header is harmless for the plain path.
+const JSON_HEADERS = { 'content-type': 'application/json' }
+
+function toContentItem(payload: unknown) {
+  return {
+    type: 'text',
+    text: typeof payload === 'string' ? payload : JSON.stringify(payload),
+  }
+}
+
 /**
- * Build a MockResponseInit whose body is a JSON-RPC envelope wrapping the given
- * content payloads. Each payload becomes one `content` item whose `text` is the
- * JSON-stringified value, mirroring how the MCP server streams tool results.
- *
- * The hooks' `mcpCall` reads the response via `response.text()` (no
- * `text/event-stream` content-type) then `JSON.parse`s it, and finally
- * `JSON.parse`s each `content[i].text`. So the hook receives the parsed
- * payloads as its result array.
+ * Build the raw JSON-RPC envelope OBJECT wrapping the given content payloads
+ * (each becomes one `content[i].text`, JSON-stringified — mirroring how the
+ * MCP server streams tool results). Use this when a test builds its own
+ * `Response`/`mockFetch` by hand and needs the body object; use `mcpResult`
+ * when you want a ready-to-route `MockResponseInit`.
+ */
+export function mcpEnvelopeJson(...payloads: unknown[]) {
+  return {
+    jsonrpc: '2.0',
+    id: 1,
+    result: { content: payloads.map(toContentItem) },
+  }
+}
+
+/**
+ * A `MockResponseInit` whose body is the JSON-RPC envelope from
+ * `mcpEnvelopeJson`. `mcpCall` finally `JSON.parse`s each `content[i].text`,
+ * so the hook receives the parsed payloads as its result array.
  */
 export function mcpResult(...payloads: unknown[]): MockResponseInit {
+  return { status: 200, headers: JSON_HEADERS, json: mcpEnvelopeJson(...payloads) }
+}
+
+/** Raw tool-error envelope object (result.isError = true). */
+export function mcpToolErrorJson(message: string) {
   return {
-    status: 200,
-    json: {
-      jsonrpc: '2.0',
-      id: 1,
-      result: {
-        content: payloads.map((p) => ({
-          type: 'text',
-          text: typeof p === 'string' ? p : JSON.stringify(p),
-        })),
-      },
-    },
+    jsonrpc: '2.0',
+    id: 1,
+    result: { isError: true, content: [{ type: 'text', text: message }] },
   }
 }
 
 /** Build a JSON-RPC tool-error envelope (result.isError = true). */
 export function mcpToolError(message: string): MockResponseInit {
-  return {
-    status: 200,
-    json: {
-      jsonrpc: '2.0',
-      id: 1,
-      result: {
-        isError: true,
-        content: [{ type: 'text', text: message }],
-      },
-    },
-  }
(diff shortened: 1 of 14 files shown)


frontend/src/hooks/mcpEnvelope.testhelper.ts
```diff
@@ -1,56 +1,64 @@
...
+/**
+ * A `MockResponseInit` whose body is the JSON-RPC envelope from
+ * `mcpEnvelopeJson`. `mcpCall` finally `JSON.parse`s each `content[i].text`,
+ * so the hook receives the parsed payloads as its result array.
  */
 export function mcpResult(...payloads: unknown[]): MockResponseInit {
+  return { status: 200, headers: JSON_HEADERS, json: mcpEnvelopeJson(...payloads) }
+}
+
+/** Raw tool-error envelope object (result.isError = true). */
```
mruwnik, line 40: **Dors** *(Claude Opus 4.6)*: [SUGGESTION] `mcpToolErrorJson` (here) and `mcpRpcErrorJson` (line 54) appear to have no consumers. Since a newly-added export can only be referenced by a file that this PR also modifies, and none of the 13 migrated files import either symbol, these two raw variants are unused dead code. `mcpEnvelopeJson` is genuinely needed (Search/SearchForm/Polls/BooksPanel/useDiscord build their own `Response`), but no test currently hand-builds a tool-error or rpc-error Response — those paths all use the `MockResponseInit` forms (`mcpToolError`/`mcpRpcError`).

Two reasonable resolutions:
- Drop the two unused raw variants (YAGNI — add them when a call site actually needs them), or
- Keep them deliberately for API symmetry (envelope/toolError/rpcError × raw/MockResponseInit) and say so — but then they're untested, unused surface.

70% I'd lean toward dropping them; the symmetry argument is defensible but weak for a test helper. Non-blocking either way.

frontend/src/hooks/useDiscord.test.ts
```diff
@@ -2,18 +2,15 @@ import { describe, it, expect, beforeEach } from 'vitest'
...
 import { useDiscord } from './useDiscord'
 import { setAuthCookies, clearCookies, mockFetch, mockResponse } from '@/test/utils'
+import { mcpEnvelopeJson } from './mcpEnvelope.testhelper'
 
-// Build the JSON-RPC tools/call body that useMCP.mcpCall expects:
-// { result: { content: [ { text: JSON.stringify(payload) } ] } }
-function mcpResult(payload: unknown) {
+// Build a full Response wrapping the JSON-RPC tools/call envelope that
+// useMCP.mcpCall expects (single payload → one content item).
+function mcpResponse(payload: unknown) {
```
mruwnik, line 9: **Dors** *(Claude Opus 4.6)*: [NITPICK] This local `mcpResponse` wrapper could be eliminated entirely in favor of the pattern `useCalendar.test.ts` already uses at its `routeMcpDynamic`: `mockResponse(mcpResult(payload))`. `mcpResult` is variadic, so a single object payload works (`mcpResult(payload)`), and `mockResponse` accepts the `MockResponseInit` it returns (incl. the `content-type` header). That would fully deliver this PR's stated goal — "no inline envelope definitions remain" — for useDiscord too, and the call sites become `mockFetch(async () => mockResponse(mcpResult({ channels: [...] })))`. Keeping the thin wrapper is fine if you prefer the local name; just flagging the asymmetry with useCalendar. Non-blocking.

mruwnik (commented):
**Dors** *(Claude Opus 4.6)*: ## Review Summary

**Approve.** This is a clean, low-risk, well-documented test-only refactor that delivers exactly what #89 asked for: the 13 inline `mcpEnvelope`/`mcpResult` copies are gone, centralized on `mcpEnvelope.testhelper.ts`. CI is green (backend + frontend). No blocking or important issues.

I verified the parts most likely to hide a regression:

- **Array→variadic conversion is correct per call site.** `mcpEnvelope([schemas])` → `mcpEnvelopeJson(schemas)` (one content item) and `mcpEnvelope(opts.results ?? [])` → `mcpEnvelopeJson(...(opts.results ?? []))` (N content items) in Search/SearchForm both preserve the original content-item count.
- **No behavior change in payload encoding.** `toContentItem` keeps the `typeof === 'string' ? p : JSON.stringify(p)` semantics from the prior shared helper; none of the migrated call sites pass a bare string, so the path is moot here.
- **Header addition is sound.** Adding `content-type: application/json` to `mcpResult`/`mcpToolError`/`mcpRpcError` is what lets `parseJsonResponse` consumers reuse them; harmless for the `mcpCall` text() path. Dropping the redundant `isError: false` from the old useBooks-family inline helper is harmless (absent ⇒ falsy).
- **useDiscord rename is complete** — all `mcpResult` call sites became `mcpResponse`, no stragglers, no shadowing of the shared name.
- **useCalendar correctly dropped its now-unused `MockResponseInit` import** (it switched to `mockResponse(mcpResult(...))`); the other hook tests kept the import because `routeMcp(resp: MockResponseInit)` still uses it.

## Blocking Issues
- None.

## Important Issues
- None.

## Suggestions
- `mcpToolErrorJson` / `mcpRpcErrorJson` raw variants have no consumers in the PR (and a new export can only be referenced by a modified file) — likely YAGNI dead code; drop or justify the API-symmetry intent.
- (nit) useDiscord's local `mcpResponse` wrapper could be replaced with the `mockResponse(mcpResult(...))` pattern useCalendar already uses, removing the last local envelope helper.

## What Works Well
- Meticulous per-call-site handling of the array-vs-variadic mismatch — the easiest place to silently break content-item counts, and it's correct everywhere.
- Good docstrings explaining the raw-vs-MockResponseInit split and the content-type rationale; comments describe current constraints, not history.
- PR body's per-shape migration table made verification fast and matches the actual diff.

PR #91, opened 2026-05-31
feat(search): unify filter translation into one declarative registry

mruwnik:
## Summary

The same `MCPSearchFilters` were translated to backend queries in three hand-maintained places — list/count SQL (`apply_item_filters`), BM25 SQL (`search_bm25`), and Qdrant payload (`search_chunks`) — that had silently drifted. A filter implemented in one arm but skipped in another produced confidently-wrong results (e.g. `recipients` honored by Qdrant but ignored by BM25, leaking non-matching mail through the RRF merge; `count_items` and `list_items` disagreeing on totals for the same filter set).

This introduces `src/memory/api/search/filters.py` as the single source of truth and rewires all three backends to consume it. Drift is now structurally prevented:
- `FILTER_REGISTRY` — declarative `FilterSpec`s (key → model.column, SQL op, Qdrant op).
- `SPECIAL_FILTER_KEYS` — backend-specific hand-coded filters (access/person/source_ids/confidence/observation/created_at).
- `reject_unknown_filter_keys` — one loud failure replacing the three backends' ad-hoc ValueError / warn-and-ignore / silent-skip behaviours.
- `check_registry_completeness()` — asserted at import **and** unit-tested, so adding an `MCPSearchFilters` key forces wiring it everywhere or the build fails.

## Bugs fixed while building it
- **Cartesian product**: JTI subclass joins used the polymorphic ORM entity, which re-aliased `source_item`. Now join `model.__table__` on `SourceItem.id`.
- **BM25 cross-join / access bypass**: `needs_source_join` used truthiness, so `min_size=0` skipped the SourceItem join while the registry still emitted `size >= 0` → uncorrelated cross join that also escaped the access filter. Now gated on `is_empty_value`.
- **Silent drops**: `subject`/`recipients`/`sent_at` were ignored on the SQL arms; now applied. `subject` over Qdrant raises (no faithful equivalent) rather than leaking.
- **observation_types/min_confidences** now rejected by list/count (observation-search-only).

## Known limitation
`created_at` is accepted but applied only on the SQL arm — the Qdrant chunk payload carries no created_at key, so a range there would match nothing. Documented inline; closing it needs a payload field + reindex (out of scope).

## Tests
221 passed / 28 skipped across the four affected test files; pyright clean (0 errors) on all 7 changed files.

### DIFF
--- a/src/memory/api/MCP/servers/core.py
+++ b/src/memory/api/MCP/servers/core.py
@@ -171,6 +171,56 @@ def _build_filters_section() -> str:
     return "\n".join(lines)
 
 
+def _build_list_items_description() -> str:
+    """Build dynamic description for the list_items tool.
+
+    Reuses :func:`_build_filters_section` (driven by SEARCH_FILTERS) so the
+    accepted content-metadata filters are documented identically to
+    search_knowledge_base instead of being hand-maintained. Mirrors how
+    _build_search_description injects the same section.
+    """
+    return textwrap.dedent(
+        """
+        List items without semantic search - for systematic enumeration.
+        Use for reviewing all items matching criteria, not finding best matches.
+
+        Args:
+            modalities: Filter by type: mail, blog, book, forum, photo, comic, etc. (empty = all)
+            {filters_section}
+            Observation-only filters (observation_types, min_confidences) are
+            rejected here - use search_observations for those.
+            limit: Max results per page (default 50, max 200)
+            offset: Skip first N results for pagination
+            sort_by: Sort field - "inserted_at", "size", or "id" (default: inserted_at)
+            sort_order: "asc" or "desc" (default: desc)
+            include_metadata: Include full as_payload() metadata (default True)
+
+        Returns: {{items: [...], total: int, has_more: bool}}"""
+    ).format(filters_section=_build_filters_section())
+
+
+def _build_count_items_description() -> str:
+    """Build dynamic description for the count_items tool.
+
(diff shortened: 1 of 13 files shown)


src/memory/api/MCP/servers/core.py
```diff
@@ -741,8 +800,11 @@ async def list_items(
     Use for reviewing all items matching criteria, not finding best matches.
 
     Args:
-        modalities: Filter by type: email, blog, book, forum, photo, comic, etc. (empty = all)
-        filters: Same filters as search_knowledge_base (tags, min_size, max_size, etc.)
+        modalities: Filter by type: mail, blog, book, forum, photo, comic, etc. (empty = all)
+        filters: Same content-metadata filters as search_knowledge_base (tags,
```
mruwnik: can this docstring be autogenerated from filters, same as how the search (or was it metadata) tool does?

mruwnik: **claude** *(Claude Opus 4.8)*: Yes — done in f1ddbdf. `list_items` and `count_items` now pass `description=_build_list_items_description()` / `_build_count_items_description()` to `@core_mcp.tool(...)`, and both build their filter section via `_build_filters_section()` (driven by `SEARCH_FILTERS`) — the exact same helper `_build_search_description()` uses for `search_knowledge_base`. So the documented filter set can no longer drift from the actual one. The hand-written filter list in the docstrings is gone; only the non-filter prose (observation-only-filter rejection note, return shapes) stays inline.

PR #95, opened 2026-06-05
feat(access): "hidden" sensitivity tombstone — exclude content from search/visibility

mruwnik:
## Goal
Let an admin make an email account's content invisible to search and fetch, reversibly, by setting the account's `sensitivity` to a new tombstone value `"hidden"`. The existing access-control reconciliation pipeline propagates the change to all existing messages + their Qdrant payloads for free.

`"hidden"` is a storable sensitivity that is **not** a rung on the role/sensitivity ladder: it's kept out of `SensitivityLevel` / `ALLOWED_SENSITIVITIES` (so generic content-create paths can't set it), settable only via the email-account API, and **actively excluded before every admin bypass** — so even a superadmin can't see it. Reversible: messages inherit `sensitivity` from the account, so `basic → hidden → basic` round-trips via re-inheritance.

## Enforcement ("always-on, even for admins")
Mirrors the existing always-on Qdrant `must_not`. Applied before each admin short-circuit:
- `access_control.apply_access_filter_to_query` — `sensitivity != "hidden"` before the `access_filter=None` early-out (BM25, final-merge, list/count).
- `embeddings.search_chunks` — always-present `must_not` on `sensitivity == "hidden"`.
- `access_control.user_can_access` — deny before the admin check (read guard only; edit/delete stay open so admins can un-hide/remove).
- `access_control.get_accessible_source_item_by_filename` — deny before admin short-circuit (file/attachment serving).
- `MCP core.fetch` — `sensitivity != "hidden"` in the fetch-by-id base query (`id=` + `ids=`).
- `MCP journal.can_access_journal_target` — deny hidden `source_item` targets before the admin bypass (`list_all` read + `add` write).
- `MCP deadlines.accessible_source_items` — exclude hidden in the base query.

`search_bm25` now joins `SourceItem` unconditionally and always calls the access helper (the join is required for the tombstone filter on the admin path); this also removes the old `needs_source_join` / `min_size=0` cross-join hazard.

## Storage / API / frontend
- `source_item` + `email_accounts` CheckConstraints widened to 5 values (model + Alembic migration `20260605_hidden_sensitivity`). The other 7 sensitivity constraints stay 4-valued — only email accounts can mint hidden content today (documented extension point).
- Email-account `Literal` + TS unions widened; `EmailPanel` gains a "Hidden (excluded from search)" option.

## Review history
Hardened over 3 rounds of `/review-loop` (reviewer "Usul"). All seven enforcement surfaces above were closed in response — the recurring class was "admin short-circuit before the hidden check". All review comments resolved.

## Known limitation (pre-existing, not closed here)
Mail ingested **after** an account is hidden stays at the inherited default until the next recent-tier access-control reconciliation runs — the same eventual-consistency window the whole inherited-AC system has for every sensitivity level. If `"hidden"` must be a hard guarantee for new mail, resolving inheritance at ingest is a follow-up.

## Tests
Tombstone exclusion verified end-to-end at every closed surface (helper admin path, `search_bm25` admin path, `search_chunks` must_not, `user_can_access`, file-serving, `core.fetch`, journal `list_all`/`add`, deadlines) plus DB-constraint and schema-validation tests. **680 passed** in the combined sweep (`--run-slow`, qdrant-marked excluded — no container runtime).

### DIFF
--- a/db/migrations/versions/20260605_hidden_sensitivity.py
+++ b/db/migrations/versions/20260605_hidden_sensitivity.py
@@ -1,0 +1,51 @@
+"""hidden sensitivity tombstone level
+
+Adds 'hidden' as an allowed sensitivity value on the two tables that can
+produce hidden content today: source_item (all content types, via joined-table
+inheritance) and email_accounts (the only API surface that can set it). The
+search/visibility layer excludes 'hidden' for everyone, including admins.
+
+The other 7 sensitivity check constraints (people, deadlines, ...) are left
+4-valued on purpose: only email accounts can mint hidden content for now. To
+extend hiding to another source type, add 'hidden' to its constraint the same
+way.
+
+Revision ID: 20260605_hidden_sensitivity
+Revises: 549f6728ccfd
+Create Date: 2026-06-05
+
+"""
+from typing import Sequence, Union
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "20260605_hidden_sensitivity"
+down_revision: Union[str, None] = "549f6728ccfd"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+# (table, constraint name, allowed-value tuple)
+_CONSTRAINTS = (
+    ("source_item", "valid_sensitivity_level"),
+    ("email_accounts", "valid_email_account_sensitivity"),
+)
+_FOUR = "sensitivity IN ('public', 'basic', 'internal', 'confidential')"
+_FIVE = "sensitivity IN ('public', 'basic', 'internal', 'confidential', 'hidden')"
+
+
+def _recreate(condition: str) -> None:
+    for table, name in _CONSTRAINTS:
+        op.drop_constraint(name, table, type_="check")
+        op.create_check_constraint(name, table, condition)
+
+
+def upgrade() -> None:
+    _recreate(_FIVE)
+
+
+def downgrade() -> None:
+    # Downgrade is lossy if any hidden rows exist — they would violate the
+    # 4-value constraint. Operators must re-classify hidden content first.
+    _recreate(_FOUR)
--- a/db/migrations/versions/20260605_hidden_sensitivity.py
+++ b/db/migrations/versions/20260605_hidden_sensitivity.py
@@ -5,10 +5,8 @@ produce hidden content today: source_item (all content types, via joined-table
 inheritance) and email_accounts (the only API surface that can set it). The
 search/visibility layer excludes 'hidden' for everyone, including admins.
 
-The other 7 sensitivity check constraints (people, deadlines, ...) are left
-4-valued on purpose: only email accounts can mint hidden content for now. To
-extend hiding to another source type, add 'hidden' to its constraint the same
-way.
+Downgrade is lossy if any hidden rows exist — they would violate the 4-value
+constraint. Operators must re-classify hidden content first.
 
 Revision ID: 20260605_hidden_sensitivity
 Revises: 549f6728ccfd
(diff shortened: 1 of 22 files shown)


db/migrations/versions/20260605_hidden_sensitivity.py
```diff
@@ -0,0 +1,51 @@
...
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "20260605_hidden_sensitivity"
+down_revision: Union[str, None] = "549f6728ccfd"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+# (table, constraint name, allowed-value tuple)
```
mruwnik: this seems overly complicated for two lines :D

src/memory/api/MCP/servers/deadlines.py
```diff
@@ -558,7 +558,14 @@ def accessible_source_items(
     if not source_item_ids:
         return []
 
-    items = session.query(SourceItem).filter(SourceItem.id.in_(source_item_ids)).all()
+    # "hidden" tombstone excluded for EVERYONE (incl. admins) in the base query,
+    # so neither the returned items nor the accessible/denied counts leak the
+    # existence of hidden content. Mirrors the search / fetch exclusions.
+    items = (
```
mruwnik: wait, this seems like a potential massive security issue? Why isn't it also using the other gates? There should be a single function somewhere which is "fetch these items for this user" which makes sure that all permissions are respected. Doesn't the normal fetch item/s tool do this?

src/memory/api/email_accounts.py
```diff
@@ -141,7 +145,9 @@ class EmailAccountUpdate(BaseModel):
     send_enabled: bool | None = None
     # Access control
     project_id: int | None = None
-    sensitivity: Literal["public", "basic", "internal", "confidential"] | None = None
+    # See EmailAccountCreate.sensitivity — "hidden" tombstones the account's
+    # mail. Reversible: messages re-inherit a normal level when changed back.
+    sensitivity: Literal["public", "basic", "internal", "confidential", "hidden"] | None = None
```
mruwnik: isn't there a mixin in source_items.py that adds these fields? That one should be used everywhere and updated appropriately

db/migrations/versions/20260605_hidden_sensitivity.py
mruwnik: **Daniel** *(Claude Opus 4.8 (1M context))*: Simplified — the migration is now four inline `op.drop_constraint` / `op.create_check_constraint` calls in upgrade/downgrade (two constants for the 4- and 5-value SQL), no `_recreate`/`_CONSTRAINTS` helper. 😄

src/memory/api/MCP/servers/deadlines.py
mruwnik: **Daniel** *(Claude Opus 4.8 (1M context))*: You're right — consolidated. There's now a single choke point in `MCP/access.py`:

- `fetch_accessible_source_items(session, ids, access_filter, *, base_query=None)` — applies the caller's `AccessFilter` via the canonical `apply_access_filter_to_query` (the same gate search / BM25 / `list_items` / `count_items` use), so creator / person / public / project-role rules AND the hidden tombstone are enforced in one place.
- `build_mcp_user_access_filter(session, user)` — builds that filter for a `UserProxy` using the passed session (no nested session).

Both deadlines (`attach` + `upsert`) and `core.fetch` now go through `fetch_accessible_source_items` instead of re-deriving access. The old deadlines version with its `if has_admin_scope: return items` short-circuit + per-item `user_can_access` loop is gone, and `core.fetch` dropped its manual loop + inline `sensitivity != "hidden"`.

To be precise on the severity: the old code wasn't *unguarded* — non-admins already went through `user_can_access`, and admins are meant to see everything. The real risk you're pointing at is exactly what bit this PR three times in review: parallel access paths that drift. One function removes that.

One subtlety I preserved: `core.fetch` keeps its **token-derived** access filter (`get_current_user_access_filter`), not a user-scope recompute — that's what honours API-key scope overrides (a `[read]`-scoped admin key can't inherit `*`), covered by `test_core_access_filter`. So the helper takes a pre-built `access_filter` rather than recomputing from the user, letting each caller pick its scope source. Verified: deadlines + core_tools + core_access_filter suites green.

src/memory/api/email_accounts.py
mruwnik: **Daniel** *(Claude Opus 4.8 (1M context))*: Good push — centralized the values. There's now a single source of truth in `memory.common.scopes` (the low-level module the models already import, next to `SensitivityLevelLiteral`):

- `StorableSensitivityLiteral` = the role ladder + `"hidden"`,
- `STORABLE_SENSITIVITIES` = `get_args(...)` of it,
- `STORABLE_SENSITIVITY_CHECK_SQL` = the derived `sensitivity IN (...)` string.

`source_item` and `email_accounts` CHECK constraints now use `STORABLE_SENSITIVITY_CHECK_SQL`; the email-account API `sensitivity` fields use `StorableSensitivityLiteral`. So adding a storable value happens in exactly one place. (The derived SQL is byte-identical to the old literal, so no DB change.)

Two things I deliberately did *not* fold in, and why:

1. **"hidden" stays out of the role ladder on purpose.** It's a tombstone no role grants — that's the whole tombstone design (kept out of `SensitivityLevelLiteral` / `ROLE_SENSITIVITY` / `ALLOWED_SENSITIVITIES` so generic create paths can't set it). The email-account API is the one intentional surface that may set it, which is why that Literal is the wider `StorableSensitivityLiteral`. So there are legitimately *two* sets (role-granted vs storable), now both derived from one literal each.

2. **`AccessControlMixin` adoption.** You're right that `EmailAccount` (and the other data-source account types — Google/Calendar/Transcript) define their own `project_id`/`sensitivity` columns instead of using the mixin. Folding them onto `AccessControlMixin` is a good cleanup, but (a) it spans all the account types, not just email, so it's orthogonal to the tombstone, and (b) the mixin can't carry the CHECK constraint anyway — its own docstring notes SQLAlchemy shares a single constraint object across subclasses, so per-table constraints stay regardless. I'd rather do that as a focused follow-up than smuggle a cross-account model refactor into this PR. Happy to open one if you want it.

PR #103, opened 2026-07-02
feat(sessions): search over archived Claude session transcripts

mruwnik:
Closes #100.

Adds `claude_session_search` / `claude_session_fetch` MCP tools so archived Claude Code session transcripts (retained 2 years) become queryable instead of requiring exact session IDs.

## What's here

- **`SessionSegment` SourceItem subtype** — embedding-sized runs of conversational messages (tool traffic stripped) riding the standard hybrid pipeline (Qdrant + BM25 + RRF) in a new `session` collection, with transcript line indices pointing back into the JSONL for context windows.
- **Indexer** — hourly `index_stale_sessions` sweep + per-session `index_session` task with an `indexed_up_to` watermark. The trailing partial segment is held back until the transcript is idle (`SESSION_INDEX_MIN_IDLE_SECONDS`, default 1h) so growing sessions don't produce overlapping segments. sha256 dedup makes reindexing idempotent; an embed failure halts the watermark so the slice is retried instead of silently dropped. Existing sessions backfill automatically on first sweep.
- **Owner-only, even for admins** — segments carry `creator_id` + explicit NULL `project_id`; `session_search` prefilters `source_ids` by `Session.user_id`; and every generic surface (`search`, `list_items`, `count_items`, `fetch`, `get_metadata_schemas`, available-modalities) excludes the session modality entirely.
- **Retention** — `cleanup_old_claude_sessions` now deletes segment `source_item` parent rows (Qdrant points are GC'd by the existing clean-all-collections sweep).
- **Acceptance criterion from the issue**: "when did we discuss X?" resolves in two tool calls — `claude_session_search(query)` → `claude_session_fetch(session_id, start_index)`.

## Review

3 rounds of differ-review (all comments resolved): round 1 fixed LIKE-wildcard escaping, lean SearchConfig (no HyDE/query-analysis LLM calls), and a cheaper hourly sweep; round 2 verified fixes, no new findings; round 3 fixed the embed-failure watermark gap, the `get_metadata_schemas` session leak, and a negative-limit edge. A separate adversarial pass also closed an admin-visibility hole in `list_items`/`count_items`/`fetch`.

Tests: 63 new tests (parser, indexing tasks, MCP tools, access-control regressions); affected suites (~1640 tests) pass; pyright + ruff clean on changed files.

https://claude.ai/code/session_012YX6tYnmz3K8LZg6MXfHCv

### DIFF
--- a/src/memory/api/MCP/servers/claude.py
+++ b/src/memory/api/MCP/servers/claude.py
@@ -1,18 +1,35 @@
-"""MCP subserver for cloud-claude session file transfer.
+"""MCP subserver for Claude Code sessions.
 
-Tools here orchestrate file transfer between the user's local machine and a
-remote claude-cloud session container. Bytes don't flow through MCP; instead
-these tools mint short-lived signed URLs that a bundled bash script (in the
-session-files skill) can curl directly. This sidesteps MCP payload caps and
-gives natural streaming for arbitrary file types and folders.
+Two tool families live here:
+
+- Cloud-container file transfer (``session_list``/``session_list_dir``/
+  ``session_pull_url``/``session_push_url``): orchestrate file transfer
+  between the user's local machine and a remote claude-cloud session
+  container. Bytes don't flow through MCP; these tools mint short-lived
+  signed URLs that a bundled bash script (in the session-files skill) can
+  curl directly, sidestepping MCP payload caps.
+
+- Archived-transcript search (``session_search``/``session_fetch``): query
+  the stored JSONL transcripts of past Claude Code sessions. Strictly
+  owner-only — sessions are personal working data, so even admins only see
+  their own here.
 """
 
 import logging
+from collections import deque
+from datetime import datetime
 from typing import Any
+from uuid import UUID as PyUUID
 
 from fastmcp import FastMCP
+from sqlalchemy import or_
+from sqlalchemy.orm import joinedload
 
-from memory.api.MCP.access import get_mcp_current_user
+from memory.api.MCP.access import (
+    build_user_access_filter_from_dict,
+    get_mcp_current_user,
+    log_search_access,
+)
 from memory.api.MCP.visibility import require_scopes, visible_when
 from memory.api.cloud_claude import (
     get_user_id_from_session,
@@ -22,17 +39,30 @@ from memory.api.orchestrator_client import (
     OrchestratorError,
     get_orchestrator_client,
 )
+from memory.api.search.search import search as search_base
+from memory.api.search.types import SearchConfig, SearchFilters
 from memory.api.transfer_tokens import (
     mint_transfer_url,
     normalize_abs_path,
     validate_transfer_path,
 )
-from memory.common import settings
+from memory.common import extract, settings
+from memory.common.dates import parse_iso_datetime_utc
+from memory.common.db.connection import make_session
+from memory.common.db.models import CodingProject, Session, SessionSegment
 from memory.common.scopes import SCOPE_READ, SCOPE_WRITE
+from memory.parsers import claude_sessions
 
 logger = logging.getLogger(__name__)
 
+MAX_FETCH_MESSAGES = 100
+
 claude_mcp = FastMCP("memory-claude")
+
+
+def like_escape(value: str) -> str:
+    """Backslash-escape LIKE metacharacters (%, _, \\) for use in a pattern."""
+    return value.replace("\\", "\\\\").replace("%", "\\%").replace("_", "\\_")
 
 
 def user_owns(user_id: int, session_id: str) -> bool:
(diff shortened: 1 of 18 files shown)


src/memory/api/MCP/servers/claude.py
```diff
@@ -228,3 +258,279 @@ async def session_push_url(
...
+        if end_time:
+            query = query.filter(
+                or_(
+                    SessionSegment.start_time <= end_time,
+                    SessionSegment.start_time.is_(None),
+                )
+            )
+        if role:
+            query = query.filter(SessionSegment.roles.contains([role]))
+        return [row[0] for row in query.all()]
```
mruwnik, line 311: **Heironeous** *(Claude Opus 4.8)*: Non-blocking (scaling): `owned_segment_ids` returns *every* segment id the user owns, and that full list is passed as `source_ids` into both the Qdrant prefilter and the BM25 `IN (...)` clause. With 2-year retention a prolific user could accumulate tens of thousands of segments, at which point this becomes a large id-set match on every search (and a very large SQL `IN` list). Fine at current scale; worth keeping an eye on. If it does become a problem, the owner gate could instead be pushed into the search filter as a `session_id`/`creator_id` predicate rather than an enumerated id list. 70% confident this is only a latency concern and never a correctness one, since the owner scoping is also enforced by `creator_id` in the access filter.

mruwnik (commented):
**Heironeous** *(Claude Opus 4.8)*: Deep independent review — LGTM, no blocking issues found. This is a carefully engineered PR; the security model and the indexing state machine both hold up under scrutiny.

**Access control (the risk that matters here) is sound.** The owner-only-even-for-admins guarantee is enforced by `owned_segment_ids` scoping `source_ids` to `Session.user_id == user_id` — which holds regardless of admin `*` scope, since an admin's access_filter is `None` and would otherwise bypass everything. Defense-in-depth is real: I verified all five generic surfaces exclude `SESSION_COLLECTIONS` — `search`, `fetch`, `list_items`/`count_items` (both share `apply_item_filters`), `get_metadata_schemas`, and `_get_available_modalities`. `session_fetch` re-checks `session.user_id != user.id`. Segments carry `creator_id` + explicit NULL `project_id`, and there's no REST search route (search is MCP-only; `search_base` requires explicit modalities rather than defaulting to all), so no alternate leak path. Same "Session not found" error on missing vs. non-owned = no existence leak.

**Indexing state machine is correct.** I traced the tricky paths:
- The IntegrityError-on-concurrent-insert path is safe: `process_content_item` commits each *successful* segment before the loop advances, so a rollback of a later duplicate can't orphan an earlier committed segment or leave a watermark gap — the watermark is re-set and re-committed on the next successful segment.
- The embed-failure halt (delete FAILED row, hold `indexed_at` stale, break) correctly forces the hourly sweep to retry the slice instead of advancing past a permanent zero-chunk gap.
- Segmentation is deterministic from the watermark, and the sha256(content, session_id, start_index) + unique (session_id, start_index) index make reindex idempotent from both directions (existence check catches the common case, IntegrityError catches the race).
- Tail-holdback via `transcript_idle` + the `tail_may_be_pending` requeue condition is self-terminating (stops requeuing once mtime is >IDLE before indexed_at).

Migration, retention cascade (delete `source_item` parents before session rows so chunks/segments cascade), LIKE-escaping, and the lean SearchConfig (no HyDE/query-analysis) all check out.

**Minor / optional (none blocking):**
1. Left an inline note on `owned_segment_ids` re: unbounded `source_ids` as a future scaling concern.
2. A SKIPPED segment (too short to embed) is counted toward `created` in the result dict though it produces no chunks — purely cosmetic; it's still sha256-deduped on reindex so it won't be recreated.
3. `sessions.py::extract_conversation_text` (summarizer) and `claude_sessions.py::parse_event`/`extract_text_blocks` (indexer) are two independent implementations of "pull conversational text from a transcript event" that differ subtly (is_meta handling, per-message truncation, join char). Consider having the summarizer reuse the parser to avoid drift. DRY nit only.

I'll keep watching and re-review on changes.

src/memory/api/MCP/servers/claude.py
mruwnik: **Claude** *(Claude Fable 5)*: Agreed this is latency-only, not correctness — and I'd put the confidence higher than 70%. Owner scoping is guaranteed independently of the id list by the access filter: `owned_segment_ids` sets `creator_id=session.user_id` on every `SessionSegment` (workers/tasks/sessions.py:333), and `build_user_access_filter_from_dict` adds `creator_id == user.id` (access_control.py:269-270). So even if the enumerated `source_ids` were dropped entirely, a user could never see another user's segment.

One wrinkle for the suggested refactor: the id list isn't only an owner gate — it also carries the `project`/`start`/`end`/`role` predicates, which live on `SessionSegment`/`Session`/`CodingProject` columns and aren't filterable fields on the indexed "session" items. So pushing owner-scope into a `creator_id` search predicate removes the owner enumeration but the metadata filtering would still need some segment-level lookup. If this becomes a latency problem at scale, the cleaner fix is probably to project `session_id`/`start_time`/`end_time`/`roles` into the Qdrant payload + BM25 row so the whole filter moves into the search layer. Leaving as-is for now since it's fine at current scale.

mruwnik:
**Heironeous** *(Claude Opus 4.8)*: Re-reviewed at head fe0cfd2 (was f3a53bc at my first pass). The only delta is a test-only refactor: imports (`json`, `uuid_lib`, `SearchResult`, `settings_module`, `CodingSession`, `CodingProject`, `SessionSegment`) hoisted from inside a test body to the top of `tests/memory/api/MCP/test_claude.py` — matches the repo's imports-at-top convention. No production code changed; checks are green. Prior assessment stands: LGTM, no blocking issues.

src/memory/api/MCP/servers/claude.py
mruwnik: **Heironeous** *(Claude Opus 4.8)*: Agreed on all points — good catch that the id list is doing double duty as owner gate *and* metadata (project/date/role) filter, so my "just move owner-scope into a creator_id predicate" suggestion was incomplete. Projecting session_id/start_time/end_time/roles into the Qdrant payload + BM25 row is the right shape if this ever becomes a latency issue, since it moves the whole filter into the search layer. Fine to leave as-is at current scale. Resolving.

PR #105, opened 2026-07-07
feat(auth): dedicated MCP tool for one-time key minting, rate-limited via soft delete

mruwnik:
## Summary

- Split one-time API key minting out of `meta_get_user` into a dedicated `create_one_time_key` MCP tool (`SCOPE_WRITE`), with a per-user rolling-window mint cap (`ONE_TIME_KEY_RATE_LIMIT`, default 10/hour) enforced by counting `api_keys` rows under a per-user `pg_advisory_xact_lock`.
- Consuming a one-time key is now a **soft delete**: atomic `UPDATE ... SET revoked WHERE NOT revoked RETURNING id` (same CWE-367 race fencing as the old `DELETE`), so consumed keys stay countable inside the mint window. A daily maintenance task (`cleanup_used_one_time_keys`) purges tombstones once they age out.
- Fixed a race in `oauth_provider.lookup_principal`: it ignored `handle_api_key_use`'s return value, letting an MCP race-loser authenticate with an already-consumed key.

## Hardening from the review loop (3 rounds, all comments resolved)

- Minted one-time keys get a TTL (`ONE_TIME_KEY_TTL_SECONDS`, default 1h); unused-but-expired keys are purged once outside the mint window.
- Rate-window comparisons run on the DB clock in-query (`created_at > now() - interval`), immune to non-UTC Postgres `TimeZone` or app/DB clock skew.
- REST `POST /users/{id}/api-keys` rejects `key_type=one_time` (422 → use the MCP tool) and permanent-delete refuses one-time keys, so REST can't route around the TTL or mint limit.
- Key listings hide consumed one-time tombstones; revoked regular keys stay visible per the existing revoke contract.

Deferred (acknowledged in review): migrating `APIKey.created_at`/`expires_at` to `DateTime(timezone=True)` — the remaining naive-datetime seam is in pre-existing `is_valid()` and only matters for non-UTC DB deployments.

## Testing

- 2767 passed across the full API tree (`tests/memory/api` + user models + rate limit) against real Postgres
- New tests: rolling-window semantics (used keys still count, window slides, per-user buckets), TTL at mint, tombstone purge fates, REST rejection paths, listing filter, race-loser fencing

https://claude.ai/code/session_01EjRsK6NrPi9j8moiYHkxhX


src/memory/api/MCP/servers/meta.py
```diff
@@ -295,29 +301,84 @@ async def get_current_time() -> dict:
...
+@visible_when(require_scopes(SCOPE_WRITE))
+async def create_one_time_key() -> dict:
+    """Mint a one-time API key for client operations.
+
+    The key stops working after its first use. It carries the caller's own
+    scopes capped to the OAuth grant — it can never hold more scope than
+    the caller already had.
+
+    Minting is rate limited per user over a rolling window
+    (ONE_TIME_KEY_RATE_LIMIT, default 10/hour).
```
mruwnik: MCP callers see this as the description of the tool. They have no idea what this limit is, as they don't see the code. Remove specific values unless they can be inserted here. That being said, it's valuable info, so maybe also return the remaining number of keys in the response?

mruwnik: **Deneir** *(Claude Fable 5)*: Fixed in b158640. The description now bakes in the deployment's actual configured values at import time ("...expires unused after 3600 seconds... rate limited per user to 10/hour...") — no setting names. And the response now reports quota: `{"one_time_key": ..., "expires_at": ..., "mints_remaining": n, "rate_limit": "10/hour"}`. Tests pin both (description contains the configured values and no setting names; `mints_remaining` decrements across mints).

Also pushed 33d1052 fixing the pre-existing frontend CI red (master's 26b845a changed MCP calls to POST exactly /mcp, but ~24 test files still routed fetch mocks by /mcp/&lt;tool&gt; URL — mocks now match by JSON-RPC body tool name; 1411 passed / 0 failed locally).

PR #106, opened 2026-07-09
fix(slack): robust workspace setup (resumable wizard, user-token OAuth, scope parsing)

DouwMarx:
Found doing a full Slack workspace setup end to end against a fresh instance. Two genuine bugs blocked/degraded setup, plus a wizard dead-end.

## Bugs

1. **Bot scope in the OAuth authorize URL blocks install.** The URL sent both a bot `scope` and `user_scope`, but the backend only stores and uses the user token (`_persist_oauth_credentials` reads `authed_user`; there is no bot-token column). Slack then requires the app to have a bot user and fails install with "doesn't have a bot user to install". Fixed by requesting `user_scope` only.

2. **Granted scopes parsed on whitespace, not comma.** Slack returns `scope` comma-separated (`"channels:history,users:read"`). `.split()` collapsed it into one bogus element, so any per-scope membership check (`"channels:history" in creds.scopes`) was always false. Fixed with `.split(",")`.

## Robustness

3. **Resumable setup wizard.** A half-configured draft app was a dead end: reopening the wizard looped back through already-completed OAuth with no path to the remaining steps. Now it starts at the first unfinished step (from the `client_secret` / `authorized` / `signing_secret` flags) and re-fetches the wizard nonce on resume; the Sources panel surfaces a "Finish setup" entry for apps not yet live.

Scopes are unchanged from upstream and stay read-only (user-token ingest, never posts).

## Docs / tests

- README: a short Slack setup runbook with the gotchas above.
- Tests: authorize URL carries `user_scope` and no bot `scope`; granted scopes parse on comma.

All frontend and backend Slack tests pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,26 @@ python tools/discord_setup.py generate-invite --bot-token <your bot token>
 
 That returns an invite URL for adding the bot to your server.
 
+### Slack integration
+
+Set up through the web UI: Sources -> Slack -> Connect Workspace. Slack must reach this instance over public HTTPS for the OAuth callback and the events webhook, so `SERVER_URL` must be a public URL. For a local instance, expose it with a tunnel (e.g. Cloudflare Tunnel) and set `SERVER_URL` to the tunnel hostname.
+
+The app authenticates as a Slack **user**, not a bot. It reads only channels and DMs that user is in, and stores per-user credentials. Do not add bot scopes: requesting bot `scope` fails the install with "doesn't have a bot user to install".
+
+Steps:
+
+1. Create an app at [api.slack.com/apps](https://api.slack.com/apps) (From scratch). Note the Client ID, Client Secret, and Signing Secret from Basic Information.
+2. In the wizard, enter the Client ID, then the Client Secret. The wizard shows the exact redirect URL and events Request URL to paste into Slack.
+3. Under OAuth & Permissions, add User Token Scopes: `channels:history`, `groups:history`, `im:history`, `mpim:history`, `channels:read`, `groups:read`, `im:read`, `mpim:read`, `users:read`, `users:read.email`, `team:read`, `reactions:read`, `files:read`. The wizard requests exactly this set.
+4. Authorize the workspace.
+5. Paste the Signing Secret into the wizard.
+6. Under Event Subscriptions, set the Request URL (verified via the signing secret). Then under "Subscribe to events on behalf of users" add `message.channels`, `message.groups`, `message.im`, `message.mpim`. URL verification alone does not deliver messages; the event subscription is a separate step.
+7. Post the wizard's test token in any channel or DM you are in. Receiving that message event flips the app to `live`.
+
+Syncing runs only when `setup_state` is `live` or `degraded` and message collection is enabled. Enable collection per workspace; channels inherit the workspace setting unless overridden per channel. A large workspace backfills slowly because Slack rate-limits history reads. Ingested messages are searchable under the `message` modality.
+
+Setup state is held in the API process memory during the wizard (`_wizard_nonces`, `_test_message_tokens`), so restarting the API mid-setup drops the current nonce and test token; re-issue them from the wizard if that happens.
+
 ### Manually triggering Celery tasks
 
 ```bash
--- a/frontend/src/components/sources/panels/SlackAppWizard.tsx
+++ b/frontend/src/components/sources/panels/SlackAppWizard.tsx
@@ -36,12 +36,27 @@ const SETUP_STATE_TO_STEP: Record<string, Step> = {
   degraded: 'done',
 }
 
+// Resume an existing app at the first UNFINISHED step. Keying off setup_state
+// alone is wrong for a reopened `draft` app: the client secret may already be
+// set and OAuth already completed (e.g. reopened via "Finish setup"), so
+// restarting at client-secret re-triggers a redundant OAuth and loops the user.
+// The per-secret / auth flags say what's actually left to do.
+const resumeStep = (app: SlackAppResponse): Step => {
+  if (app.setup_state === 'live' || app.setup_state === 'degraded') return 'done'
+  if (app.setup_state === 'signing_verified') return 'test-message'
+  if (!app.client_secret_configured) return 'client-secret'
+  if (!app.is_authorized) return 'oauth'
+  if (!app.signing_secret_configured) return 'signing-secret'
+  return 'events-url'
+}
+
 const POLL_INTERVAL_MS = 3000
 
-// OAuth scopes requested for both bot and user tokens. Read-only across
-// the workspace surface (channels/groups/im/mpim history + read), users +
-// emails for mention resolution, reactions/files for the existing message
-// processing pipeline. Keep aligned with the backend's expectations.
+// User token scopes requested during OAuth (sent as `user_scope`; the app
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -188,6 +188,26 @@ python tools/discord_setup.py generate-invite --bot-token <your bot token>
...
+
+Steps:
+
+1. Create an app at [api.slack.com/apps](https://api.slack.com/apps) (From scratch). Note the Client ID, Client Secret, and Signing Secret from Basic Information.
+2. In the wizard, enter the Client ID, then the Client Secret. The wizard shows the exact redirect URL and events Request URL to paste into Slack.
+3. Under OAuth & Permissions, add User Token Scopes: `channels:history`, `groups:history`, `im:history`, `mpim:history`, `channels:read`, `groups:read`, `im:read`, `mpim:read`, `users:read`, `users:read.email`, `team:read`, `reactions:read`, `files:read`. The wizard requests exactly this set.
+4. Authorize the workspace.
+5. Paste the Signing Secret into the wizard.
+6. Under Event Subscriptions, set the Request URL (verified via the signing secret). Then under "Subscribe to events on behalf of users" add `message.channels`, `message.groups`, `message.im`, `message.mpim`. URL verification alone does not deliver messages; the event subscription is a separate step.
```
mruwnik, line 204: **Ame-no-Uzume** *(Claude Opus 4.8)*: The README here correctly says to subscribe **"on behalf of users"** with only the four `message.*` events — consistent with the user-token-only model this PR establishes. But the wizard UI (`SlackAppWizard.tsx`, the `events-url` step, unchanged by this PR) still tells the user the opposite:

```
Subscribe to bot events:
  • message.channels / groups / im / mpim
  • reaction_added / reaction_removed
  • channel_created / channel_renamed / channel_archived
```

Two concerns:

1. **"bot events" vs "on behalf of users".** Since the whole point of fix #1 is that the app has *no bot user* ("doesn't have a bot user to install"), Slack's "Subscribe to bot events" section isn't the right one — it generally requires a bot user. The README's "on behalf of users" is correct; the wizard should match, or a user following the on-screen wizard will land in the section that reproduces the bot-user problem. (~60% confident this is an actual setup snag; you completed setup end-to-end, so maybe Slack tolerates it or you added a bot user manually — worth confirming.)

2. **Event set mismatch.** The wizard lists `reaction_*` and `channel_*` events; the README lists only the four `message.*`. If reaction/channel ingestion actually needs those subscriptions, the README under-documents them; if it doesn't, the wizard over-asks. Either way the two should agree.

Reconciling the wizard text with this README section would close the loop on the user-token migration.

frontend/src/components/sources/panels/SlackAppWizard.tsx
```diff
@@ -155,6 +170,24 @@ export const SlackAppWizard = ({
...
+  useEffect(() => {
+    if (!app || nonceData) return
+    if (step !== 'oauth' && step !== 'signing-secret' && step !== 'events-url') return
+    let cancelled = false
+    wizard
+      .issueWizardNonce(app.id)
+      .then(n => {
+        if (!cancelled) setNonceData(n)
+      })
+      .catch(e => console.warn('Wizard nonce fetch failed:', e))
```
mruwnik, line 185: **Ame-no-Uzume** *(Claude Opus 4.8)*: On the resume path this is the *only* place the nonce gets fetched, and a failure is swallowed with `console.warn`. The `oauth` and `events-url` steps render nothing until `callbackUrl`/`eventsUrl` exist (`{step === 'oauth' && app && callbackUrl && ...}`), so if `issueWizardNonce` fails here the user is left staring at a wizard with a header, a progress bar, and an empty body — no error, no retry.

That's exactly the dead-end class this PR is trying to eliminate. The non-resume path surfaces this failure (`handleClientSecret` calls `handleError`); the resume path should too — `setError((e as Error).message)` in the catch, so at least the user sees why the step is blank and can reopen. Minor, but cheap and on-theme.

frontend/src/components/sources/panels/SlackPanel.tsx
```diff
@@ -152,10 +172,33 @@ export const SlackPanel = () => {
         </button>
       </div>
 
+      {unfinishedApps.length > 0 && !wizardOpen && (
```
mruwnik, line 175: **Ame-no-Uzume** *(Claude Opus 4.8)*: Two things on the "Finish setup" banner:

1. **Ownership.** `unfinishedApps` is `listApps()` filtered only by `setup_state`, not by `is_owner`. If `GET /slack/apps` returns apps the current user can see but doesn't own (multi-tenant), a non-owner gets a "Finish setup" button that walks them into `setClientSecret` / `setSigningSecret`, which should fail server-side with a permission error. If `listApps` is already owner-scoped on the backend this is moot — worth a quick confirm. If not, filter on `a.is_owner` here (the field exists on `SlackAppResponse`).

2. **Styling (nit).** This banner uses raw `border-yellow-300 bg-yellow-50 text-yellow-900` utilities inline while the rest of the panel goes through the `styles` module. Those are light-mode-only colors — if the app has a dark theme, yellow-900 text on yellow-50 won't adapt. Low priority, but it stands out against the surrounding convention.

mruwnik (commented):
**Ame-no-Uzume** *(Claude Opus 4.8)*: ## Review — fix(slack): robust workspace setup

Solid, well-scoped PR. The three fixes are individually correct and the comments/tests are unusually good (they explain the *why*, not just the *what*). No blocking issues; my findings are reconciliation and robustness polish.

**Verified correct:**
- **`user_scope`-only OAuth** — matches the backend, which reads only `authed_user` and has no bot-token column. Test asserting `user_scope` carries scopes and `scope` is absent is exactly right.
- **Comma-split of granted scopes** (`slack.py:436`) — `[s for s in ....split(",") if s]` correctly handles Slack's comma-separated `scope` and filters empties. Correct fix.
- **`resumeStep`** — keying resume off the `client_secret_configured` / `is_authorized` / `signing_secret_configured` flags rather than `setup_state` is the right call; the `draft`-with-OAuth-already-done loop it fixes is real. Step ordering is sound, and all three flags exist on `SlackAppResponse`.

**Findings (all non-blocking):**
1. *(README + wizard, medium)* README step 6 correctly says subscribe **"on behalf of users"** with four `message.*` events, but the wizard's `events-url` step (unchanged) still says **"Subscribe to bot events"** and lists extra `reaction_*`/`channel_*` events. For a no-bot-user app these contradict each other and the new user-token model — reconcile the wizard text with the README. Details inline on README:204.
2. *(SlackAppWizard.tsx:185, low)* On resume, the nonce fetch is the only one, and its failure is swallowed with `console.warn` → blank wizard body with no error/retry. Surface it via `setError` to match the non-resume path. On-theme with the PR's dead-end fix.
3. *(SlackPanel.tsx:175, low)* "Finish setup" banner isn't filtered by `is_owner` — confirm `listApps` is owner-scoped server-side, else a non-owner gets a button that fails mid-wizard. Plus the banner's inline `yellow-*` utilities are light-mode-only.

**One neutral note on severity:** I couldn't find any current consumer that does a per-scope membership check against `SlackUserCredentials.scopes` (grepped `src/`), so the comma-split fix is latent-correctness hygiene rather than an active bug fix today — the PR body's "always false" framing slightly overstates present impact. The fix is still correct and worth keeping.

I'll keep watching and re-review as the PR evolves.

